PR #20, opened 2025-02-05
12 containerization

fracotti:
1. Reviewed and updated the Dockerfile for the backend.
2. Modified docker-compose.db.yml to include a Redis service for caching.


BackEnd/Dockerfile
```diff
@@ -1,27 +1,23 @@
...
 
-RUN apk update \
-    && apk add postgresql-dev gcc python3-dev musl-dev
+RUN apk add --no-cache postgresql-dev gcc python3-dev musl-dev
 
-RUN pip install --upgrade pip
 COPY ./requirements.txt .
-RUN pip3 install -r requirements.txt
+
+RUN pip install --upgrade pip && pip install --no-cache-dir -r requirements.txt
```
v2496s, line 17: We should discuss it after scrum, as we may change pip to smth else

docker-compose.db.yml
```diff
@@ -33,8 +33,21 @@ services:
       - db
     networks:
       - forum_network
+
```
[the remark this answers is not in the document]
fracotti: It a good question i would like to ask on daily meeting, but forgot. Thanks for reminded me about this.

boghtml (approved):

BackEnd/Dockerfile
fracotti: Discussed today that pip is okay for now, maybe in future will be some "torture" for example poetry

docker-compose.db.yml
```diff
@@ -33,8 +33,21 @@ services:
...
+  redis:
+    image: redis:7
+    container_name: redis
+    command: [ "redis-server", "--appendonly", "yes" ]
+    restart: always
+    volumes:
+      - redis_data:/data
+    networks:
+      - forum_network
+
```
VovaD199, line 46: Redis can consume too much memory under high load.
Solution: Add a limit:

```
command: [ "redis-server", "--maxclients", "1000", "--maxmemory-policy", "allkeys-lru" ]
```

-maxclients 1000 - limits the number of clients to 1000.
-maxmemory-policy allkeys-lru - clears the oldest keys when memory runs out.

VovaD199 (changes requested):

docker-compose.db.yml
fracotti: Changed it, thank you for the advice)

docker-compose.redis.yml
VovaD199, line 1: Please check what's wrong with this file, because it doesn't appear for me "docker-compose.db.yml".

VovaD199 (approved):

PR #22, opened 2025-02-06
Added indexing fields in profiles

Maksym04u:
dev
## JIRA

## Code reviewers

- [ ] @github_username

### Second Level Review

- [ ] @github_username

## Summary of issue

ToDo

## Summary of change

ToDo

## Testing approach

ToDo

## CHECK LIST
- [ ]  СI passed
- [ ]  Сode coverage >=95%


BackEnd/profiles/models.py
```diff
@@ -133,6 +134,17 @@ class Profile(models.Model):
 
     objects = ProfileManager.as_manager()
 
+    class Meta:
+        indexes = [
+            models.Index(fields=('name')),
+            models.Index(fields=('official_name')),
+            models.Index(fields=('created_at')),
+            models.Index(fields=('updated_at')),
```
NikitaPash, line 142: You should probably add trailing commas for "fields" to become tuples. Right now they are treated as strings. Or you can leave it as it is if it works as intended.

NikitaPash (approved):
Other than that comment, it looks nice for me.

BackEnd/profiles/models.py
mehalyna, line 142: The current definition uses parentheses around a single string instead of a tuple. This will not work correctly because `fields=('name')` is interpreted as a string, not a tuple. Use a tuple with a trailing comma:
```
models.Index(fields=('name',)),
models.Index(fields=('official_name',)),
models.Index(fields=('created_at',)),
models.Index(fields=('updated_at',)),
```

BackEnd/profiles/models.py
```diff
@@ -133,6 +134,17 @@ class Profile(models.Model):
...
+            GinIndex(name="service_info_gin", fields=["service_info"]),  # using in advanced search
+            GinIndex(name="product_info_gin", fields=["product_info"]), # using in advanced search
+            GinIndex(name="common_info_gin", fields=["common_info"])   # using in advanced search
```
mehalyna, line 145: `GinIndex` is valid only for PostgreSQL, and it works best on `JSONField`, `ArrayField`, or `TextField` with a gin index.
Ensure your `service_info`, `product_info`, and `common_info` fields are compatible (`TextField` is fine, but confirm if PostgreSQL pg_trgm extension is enabled for full-text search).

BackEnd/profiles/models.py
```diff
@@ -133,6 +134,17 @@ class Profile(models.Model):
...
+        ]
+
```
mehalyna, line 147: Currently, there is no default ordering, which can lead to unpredictable query results. Consider adding a sensible default ordering.

mehalyna, line 147: If profiles should not have duplicate `name` and `official_name`, enforce uniqueness.

mehalyna (changes requested):

BackEnd/profiles/models.py
Maksym04u: I am not sure if we need to add uniqueness to these fields, because profile is kinda complex model. It can represent investors, startups. So i suppose there moght be a chance for duplicates.

BackEnd/profiles/models.py
Maksym04u: Yeah my bad. Already changed it

BackEnd/profiles/models.py
Maksym04u: I suggest to add ordering by date. For me seems the most reasonable.

mehalyna (approved):

Issue #6, opened 2025-02-04
Project Structure Improvement


mehalyna:
## **Task 1: Project Structure Improvement**
- **Objective:** Restructure the backend to adhere to clean architecture principles.
- **Subtasks:**
  - Create a `services` directory for business logic.
  - Separate the models, serializers, and views into dedicated submodules.
  - Ensure the `utils` folder contains only reusable utility functions.
  - Clean up unused or redundant files and folders.

PR #25, opened 2025-02-08
Project structure improvement #6

hrybvalerii:
dev
## JIRA

## Code reviewers

- [ ] @github_username

### Second Level Review

- [ ] @github_username

## Summary of issue

ToDo

## Summary of change

ToDo

## Testing approach

ToDo

## CHECK LIST
- [ ]  СI passed
- [ ]  Сode coverage >=95%


BackEnd/authentication/managers.py
mehalyna, line 1: The `create_user` and `create_superuser` methods don't check if the required fields (like `email` and `password`) are provided. If these fields are missing, the code might break later in the process.

BackEnd/authentication/managers.py
```diff
@@ -0,0 +1,18 @@
...
+        user.save()
+        return user
+
+    def create_superuser(self, email, password=None, **extra_fields):
+        user = self.create_user(email, password=password, **extra_fields)
+        user.is_active = True
+        user.is_staff = True
+        user.is_superuser = True
+        user.save()
+        return user
```
mehalyna, line 18: The method assumes that `is_staff` and `is_superuser` will always be set correctly. If they are not, the created superuser may not have the required privileges.
Use `extra_fields.setdefault()` to ensure the correct default values.

BackEnd/authentication/managers.py
mehalyna, line 1: The `save()` method should specify the database to ensure compatibility in multi-database setups.
Add `using=self._db` to the `save()` call.

mehalyna, line 1: The methods lack docstrings, which can make the code less understandable for new developers.

BackEnd/profiles/managers.py
```diff
@@ -1,6 +1,17 @@
...
+    
+
+    def soft_delete(self, user):
+        self.is_deleted = True
+        self.save()
+        user.is_active = False
+        user.email = (
+            f"is_deleted_{now().strftime('%Y%m%d%H%M%S')}_{user.email}"
+        )
+        user.save()
```
mehalyna, line 17: `self.is_deleted = True` applies to the entire queryset, not just a specific `Profile` instance. This is because `soft_delete` is being called on a queryset, not an individual object.
Use a method that applies to a single `Profile` instance, not the queryset. Move the `soft_delete` logic to the `Profile` model or work with individual instances.

BackEnd/profiles/views.py
```diff
@@ -207,13 +207,14 @@ def perform_destroy(self, instance):
...
-            )
-            user.save()
+            instance.objects.soft_delete(user)
+            # instance.is_deleted = True
+            # instance.save()
+            # user.is_active = False
+            # user.email = (
+            #     f"is_deleted_{now().strftime('%Y%m%d%H%M%S')}_{user.email}"
+            # )
+            # user.save()
```
mehalyna, line 217: Remove commented lines of the code before pushing to repo, pls.

mehalyna (changes requested):

BackEnd/authentication/managers.py
hrybvalerii: But is_staff and is_superuser are explicitly set to True after the call of create_user method, so it shouldn't be able to override those attributes. Am I right?

BackEnd/authentication/managers.py
hrybvalerii: I added user.full_clean() before save so now it shouldn't be possible to create a user without required fields. Is it better to also remove the default value of password=True to ensure that it is passed to the method?

hrybvalerii: I added docstrings for create_user and create_superuser methods

mehalyna: Your approach of calling `user.full_clean()` before saving ensures that required fields are validated, which is a great improvement. Removing the default value of `password=None` in `create_user` and `create_superuser` would also be beneficial, as it would enforce explicit password input and avoid accidental omissions.
By making `password` a required parameter, you prevent potential issues where a user might be created without a password, even if it’s later set manually. This aligns with best practices for security and data integrity.
So yes, I would recommend removing the default value of `password=None` to ensure that it is explicitly provided when calling `create_user` and `create_superuser`.

BackEnd/authentication/managers.py
mehalyna: For robustness and readability, it’s still a good idea to use `extra_fields.setdefault()` in `create_superuser`. It doesn’t replace your explicit assignments but serves as an extra safeguard.

mehalyna (approved):

BackEnd/profiles/managers.py
```diff
@@ -1,6 +1,9 @@
 from django.db import models
+from django.utils.timezone import now
```
fracotti, line 2: Don't understand why we need `from django.utils.timezone import now` if it didn't used in current file.

fracotti (commented):
All except `from django.utils.timezone import now` in `BackEnd/profiles/managers.py` look good to me. You did a great job!

IraMladanovych02 (approved):
Good job!

BackEnd/profiles/managers.py
hrybvalerii: It isn't needed here after I moved the deletion logic to another module, so I'll just remove this line

BackEnd/authentication/managers.py
hrybvalerii: Thanks for the feedback! 
I updated the PR accordingly.

PR #61, opened 2025-02-14
Implement user registration

sofiahnatiuk:
1. Implemented serializers.py, view.py and urls.py for registration
2. Implemented tests for registration
3. Added logging and swagger documentation for registration

### DIFF
--- a/BackEnd/authentication/serializers.py
+++ b/BackEnd/authentication/serializers.py
@@ -49,7 +49,7 @@ class CustomProfileSerializer(serializers.ModelSerializer):
         fields = ("name", "is_registered", "is_startup", "is_fop")
 
 
-class UserRegistrationSerializer(UserCreatePasswordRetypeSerializer):
+class UserRegistrationSerializer(serializers.ModelSerializer):
     company = CustomProfileSerializer(write_only=True)
     email = serializers.EmailField(
         write_only=True,
@@ -57,19 +57,20 @@ class UserRegistrationSerializer(UserCreatePasswordRetypeSerializer):
     password = serializers.CharField(
         style={"input_type": "password"}, write_only=True
     )
+    re_password = serializers.CharField(write_only=True)
     captcha = serializers.CharField(
         write_only=True, allow_blank=True, allow_null=True
     )
 
     class Meta(UserCreatePasswordRetypeSerializer.Meta):
         model = User
-        fields = ("email", "password", "name", "surname", "company", "captcha")
+        fields = ("email", "password", "re_password", "name", "surname", "company", "captcha")
 
     def validate(self, value):
         custom_errors = defaultdict(list)
         captcha_token = value.get("captcha")
         self.fields.pop("re_password", None)
-        re_password = value.pop("re_password")
+        re_password = value.get("re_password")
         email = value.get("email").lower()
         password = value.get("password")
         company_data = value.get("company")
@@ -95,7 +96,7 @@ class UserRegistrationSerializer(UserCreatePasswordRetypeSerializer):
             validate_password_strength(password)
         except ValidationError as error:
             custom_errors["password"].append(error.message)
-        if value["password"] != re_password:
+        if value['password'] != value['re_password']:
             custom_errors["password"].append("Passwords don't match.")
         if captcha_token and not verify_recaptcha(captcha_token):
             custom_errors["captcha"].append(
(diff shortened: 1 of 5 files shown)


BackEnd/authentication/tests/test_user_registration.py
NikitaPash, line 1: You can define the registration url (url = reverse('authentication:register')) only once at the top of the file, rename it to 
something like REGISTER_URL and reuse, when needed

sofiahnatiuk: Updated test_user_registration.py

BackEnd/authentication/views.py
NikitaPash, line 1: We alreade have drf-spectacular in the project, and as far as I know it can be used in the same way you used drf-yasg here

sofiahnatiuk: Updated the code

IraMladanovych02 (approved):
Good job!

BackEnd/authentication/serializers.py
```diff
@@ -95,7 +97,7 @@ def validate(self, value):
             validate_password_strength(password)
         except ValidationError as error:
             custom_errors["password"].append(error.message)
-        if value["password"] != re_password:
+        if value['password'] != value['re_password']:
```
mehalyna, line 100: Accessing `value["password"]` or `value["re_password"]` directly without a default or existence check can raise a `KeyError`.

BackEnd/authentication/tests/test_user_registration.py
mehalyna, line 1: Repeated use of the same test data (e.g.,"email": "jane@test.com", "password": "Test12!34", "re_password": "Test12!34",) across multiple tests can cause failures in cases like duplicate email validation.
The code directly creates users for validation (e.g., checking duplicate emails) instead of using a factory like `UserFactory`, which would make test setup more consistent and reusable.

BackEnd/authentication/views.py
```diff
@@ -1,6 +1,52 @@
...
+    )
+    def post(self, request):
+        serializer = UserRegistrationSerializer(data=request.data)
+        if serializer.is_valid():
+            user = serializer.save()
+            user_data = UserRegistrationSerializer(user).data
+            logger.info("User created successfully")
+            return Response(user_data, status=status.HTTP_201_CREATED)
+        logger.error(f"Unexpected error occurred: {serializer.errors}")
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna, line 52: The `UserRegistrationView` endpoint does not include rate limiting, which could make it vulnerable to brute force attacks.

BackEnd/authentication/views.py
```diff
@@ -1,6 +1,52 @@
...
+        },
+    )
+    def post(self, request):
+        serializer = UserRegistrationSerializer(data=request.data)
+        if serializer.is_valid():
+            user = serializer.save()
+            user_data = UserRegistrationSerializer(user).data
+            logger.info("User created successfully")
+            return Response(user_data, status=status.HTTP_201_CREATED)
+        logger.error(f"Unexpected error occurred: {serializer.errors}")
```
mehalyna, line 51: Logging `serializer.errors` can expose sensitive information about the request in logs. This should be sanitized or limited in production.

BackEnd/authentication/views.py
mehalyna, line 1: If an unexpected error occurs (e.g., database connection issues), the view does not catch exceptions and could crash the application.

BackEnd/authentication/views.py
mehalyna, line 52: Using the same serializer (`UserRegistrationSerializer`) for both validation and response can include unnecessary fields (e.g., `password`, `re_password`, `captcha`).

mehalyna (changes requested):

NikitaPash (approved):
Don't see any more problems except those pointed out by the mentor

AntonKovalenk0 (approved):
Agree with mentor

BackEnd/authentication/serializers.py
sofiahnatiuk: Updated serializers.py

BackEnd/authentication/tests/test_user_registration.py
sofiahnatiuk: Updated test_user_registration.py

BackEnd/authentication/views.py
sofiahnatiuk: Added rate limiting

BackEnd/authentication/views.py
sofiahnatiuk: Removed error logging

BackEnd/authentication/views.py
sofiahnatiuk: Added exception handling

BackEnd/authentication/views.py
sofiahnatiuk: Divided validation and response to remove unnecessary responce fields

mehalyna (approved):

PR #62, opened 2025-02-18
Implement user password reset and change.

IraMladanovych02:
This PR introduces the functionality for users to reset and change their passwords securely. The implementation includes:
1. Password Reset - Allows users to request a password reset via email and confirm it using a secure token:
- Users can request a password reset by providing their email address.
- A secure token is generated and sent to the user's email.
- Users can confirm the reset by providing the token and a new password.
- Token validation ensures security and prevents unauthorized access.
2. Password Change - Enables authenticated users to change their password after verifying their current password:
- Authenticated users can change their password by providing their current password and a new password.
- The new password is validated for strength and complexity.
- The user's session remains active after the password change


sofiahnatiuk (approved):

BackEnd/authentication/views.py
mehalyna, line 1: The API leaks whether an email exists in the system. Attackers can check if an email is registered by brute-force guessing emails.
You may always return a 200 OK response even for non-existent emails.

BackEnd/authentication/views.py
```diff
@@ -99,3 +109,183 @@ def post(self, request):
...
+                }
+            ),
+        }
+    )
+    def post(self, request):
+        token = request.data.get("token")
+        new_password = request.data.get("new_password")
+
+        try:
+            user_id = signer.unsign(token, max_age=3600)
```
mehalyna, line 232: The token is valid indefinitely unless expired in `signer.unsign(token, max_age=3600)`. Attackers with leaked reset links can reset passwords long after a request.
Ensure tokens expire and invalidate used tokens.

BackEnd/authentication/views.py
```diff
@@ -99,3 +109,183 @@ def post(self, request):
...
+                }
+            ),
+        }
+    )
+    def post(self, request):
+        serializer = PasswordChangeSerializer(data=request.data, context={'request': request})
+        if serializer.is_valid():
+            serializer.save()
+            return Response({"message": "Password was updated successfully."}, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
mehalyna, line 291: Users remain logged in after changing their password. If a session is hijacked, the attacker can stay logged in even after the victim changes their password.
Force a session logout on password change (invalidate all active sessions).

mehalyna (changes requested):

Maksym04u (approved):

mehalyna (approved):

PR #63, opened 2025-02-18
#task53 account activation

sofiahnatiuk:
1. Implemented account validation
2. A token is generated for each new user
3. New users get an email with an activation link
4. If the token is valid, user is marked as active
5. All tests in test_user_activation.py passed successfully


BackEnd/authentication/serializers.py
```diff
@@ -230,3 +230,24 @@ def validate(self, data):
...
+    email = serializers.EmailField()
+    def validate_email(self, value):
+        try:
+            user = User.objects.get(email=value)
+        except User.DoesNotExist:
+            raise serializers.ValidationError("User with this email does not exist.")
+            logger.error("User with this email does not exist.")
+        if user.is_active:
+            raise serializers.ValidationError("This account is already active.")
+            logger.error("This account is already active.")
```
IraMladanovych02, line 245: The error message "This account is already active." exposes whether an email exists in the database. This can be exploited by attackers to check for registered emails.

BackEnd/authentication/views.py
```diff
@@ -70,13 +79,86 @@ def post(self, request):
...
+
+        if not token:
+            logger.error(f"No token provided.")
+            return Response({"detail": "No token provided."}, status=status.HTTP_400_BAD_REQUEST)
+
+        signer = TimestampSigner()
+        try:
+            uid = signer.unsign(token, max_age=3600)
+
+            user = get_user_model().objects.get(pk=uid)
```
IraMladanovych02, line 147: please, add User = get_user_model() at the beginning of the module. So, you can use user = User.objects.get(pk=uid)

BackEnd/authentication/views.py
```diff
@@ -70,13 +79,86 @@ def post(self, request):
...
+
+            user.is_active = True
+            user.save()
+
+            logger.info("Account activated successfully.")
+            return Response({"detail": "Account activated successfully."}, status=status.HTTP_200_OK)
+        except BadSignature:
+            logger.error(f"Invalid or expired token.")
+            return Response({"error": "Invalid or expired token."}, status=status.HTTP_400_BAD_REQUEST)
+        except get_user_model().DoesNotExist:
```
IraMladanovych02, line 157: please, use the same structure here - except User.DoesNotExist:

BackEnd/authentication/serializers.py
sofiahnatiuk: Changed the error message

IraMladanovych02 (approved):
good job!

BackEnd/authentication/views.py
```diff
@@ -70,13 +79,87 @@ def post(self, request):
                 user = serializer.save()
                 user_data = UserRegistrationResponseSerializer(user).data
                 logger.info("User created successfully")
+
+                signer = TimestampSigner()
+                uid = str(user.pk)
+                signed_token = signer.sign(uid)
+
+                email_subject = "Account Activation"
+                activation_link = f"http://localhost:8080/auth/activate/?token={signed_token}"
```
VovaD199, line 88: Please replace the hardcoded address with this option:

from django.conf import settings
activation_link = f"{settings.FRONTEND_URL}/auth/activate/?token={signed_token}"

VovaD199 (changes requested):

VovaD199 (approved):

PR #66, opened 2025-02-21
40 api true

vladmorozov2:
Implemeted API for chats


BackEnd/chat/consumers.py
```diff
@@ -6,50 +6,51 @@
...
-            logger.warning("Received event without 'message' key")
+    # async def chat_message(self, event):
+    #     message = event.get("message")
+    #     if message:
+    #         try:
+    #             await self.send(text_data=json.dumps({"message": message}))
+    #         except Exception as e:
+    #             logger.error(f"Failed to send message: {e}")
+    #     else:
+    #         logger.warning("Received event without 'message' key")
```
mehalyna, line 55: Push commented lines of code to the repository is forbidden.

BackEnd/chat/models.py
```diff
@@ -1,59 +1,65 @@
...
+    }
 
-        self.save()
+    def clean(self):
+        if not self.text:
+            raise ValidationError("Message text cannot be empty.")
+        if len(self.text.strip()) == 0:
+            raise ValidationError("Message text cannot be only whitespace.")
+        if self.sender_id not in self.room.participant_ids:
+            raise ValidationError("Sender is not in the room")
```
mehalyna, line 65: The `check self.sender_id not in self.room.participant_ids` will fail if room is `None` (e.g., when room is not provided). This raises an `AttributeError` instead of a `ValidationError`.
Ensure room exists before checking `participant_ids`.

BackEnd/chat/models.py
mehalyna, line 1: `updated_a`t is never updated automatically. The room does not track updates correctly.
Override the `save()` method to update `updated_at`.

mehalyna, line 1: The check for existing rooms `Room.objects(participant_ids=self.participant_ids).first()` may not work as expected, because MongoDB does not index lists properly in queries. This can cause performance issues and unexpected duplicate room creation.
Instead of querying for exact matches, use aggregation to check for duplicate participant combinations.
```
existing_room = Room.objects.filter(participant_ids__in=self.participant_ids).first()
if existing_room and set(existing_room.participant_ids) == set(self.participant_ids):
    raise ValidationError("A room with the same participants already exists.")
```

BackEnd/chat/models.py
```diff
@@ -1,59 +1,65 @@
...
-            raise ValueError(
-                f"Sender ID {sender_id} is not a participant in this room"
+        self.participant_ids = sorted(self.participant_ids)
+
+        existing_room = Room.objects(
+            participant_ids=self.participant_ids
+        ).first()
+        if existing_room:
+            raise ValidationError(
+                "A room with the same participants already exists."
```
mehalyna, line 39: The check for existing rooms may not work as expected, because MongoDB does not index lists properly in queries. This can cause performance issues and unexpected duplicate room creation.
Instead of querying for exact matches, use aggregation to check for duplicate participant combinations.
```
existing_room = Room.objects.filter(participant_ids__in=self.participant_ids).first()
if existing_room and set(existing_room.participant_ids) == set(self.participant_ids):
    raise ValidationError("A room with the same participants already exists.")
```

BackEnd/chat/serializers.py
```diff
@@ -1,65 +1,90 @@
...
 
 
-class MessageSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Message
-        fields = "__all__"
+class RoomSerializer(serializers.Serializer):
+    participant_ids = serializers.ListField(child=serializers.IntegerField())
+    created_at = serializers.DateTimeField(default=timezone.now)
+    updated_at = serializers.DateTimeField(default=timezone.now)
```
mehalyna, line 11: Default values in serializers do not work like Django models. If a user manually sends a request, they could override these timestamps.
Use `read_only=True` to prevent users from modifying these fields.

BackEnd/chat/serializers.py
```diff
@@ -1,65 +1,90 @@
...
+                "participant_ids contains duplicates"
+            )
 
-logger = logging.getLogger(__name__)
+        if not all(isinstance(user_id, int) for user_id in value):
+            raise serializers.ValidationError(
+                "All participant IDs must be integers"
+            )
 
+        return sorted(value)
```
mehalyna, line 45: The method ensures uniqueness and sorting but does not check for an empty list. A room could be created without any participants, which is illogical.
Add a check to prevent empty lists.

BackEnd/chat/serializers.py
```diff
@@ -1,65 +1,90 @@
...
-            # Join room group
-            await self.channel_layer.group_add(
-                self.room_group_name, self.channel_name
-            )
-            await self.accept()
-        except Exception as e:
-            logger.error(f"Failed to join room group: {e}")
-            await self.close()
+class MessageSerializer(serializers.Serializer):
+    room = serializers.PrimaryKeyRelatedField(queryset=Room.objects.all())
```
mehalyna, line 49: `PrimaryKeyRelatedField` only works well with integer IDs. MongoDB uses `ObjectId`, not integers, so this may fail.
Use `SlugRelatedField` referencing `id` instead.

BackEnd/chat/serializers.py
```diff
@@ -1,65 +1,90 @@
...
+    sender_id = serializers.IntegerField()
+    text = serializers.CharField()
+    timestamp = serializers.DateTimeField(default=timezone.now)
```
mehalyna, line 52: Users should not be able to set `timestamp` manually. Use `read_only=True` to prevent user modifications.

BackEnd/chat/serializers.py
```diff
@@ -1,65 +1,90 @@
...
+        """
+        Ensure that the message text is not empty or just whitespace.
+        """
+        if not value or len(value.strip()) == 0:
+            raise serializers.ValidationError(
+                "Message text cannot be empty or just whitespace."
             )
-        except Exception as e:
-            logger.warning(f"Failed to discard from room group: {e}")
+        return value
```
mehalyna, line 78: Users can send very long messages, causing performance and storage issues.
Set a max length for text (e.g., 500 characters).

BackEnd/chat/serializers.py
```diff
@@ -1,65 +1,90 @@
...
-                logger.warning("Received message without 'message' key")
-        except json.JSONDecodeError:
-            logger.warning("Invalid JSON received")
-        except Exception as e:
-            logger.error(f"Error while processing received message: {e}")
+    def to_representation(self, instance):
+        """
+        Customize serialization to handle ObjectId properly.
+        """
+        representation = super().to_representation(instance)
```
mehalyna, line 84: `to_representation()` converts MongoDB `ObjectId` to JSON, but it is not optimized. The function overwrites DRF’s default behavior, which could cause issues.
Use DRF’s built-in `json_util.loads()` only for MongoDB `ObjectId` fields.

BackEnd/chat/urls.py
```diff
@@ -1,6 +1,15 @@
 from django.urls import path
+from .views import CreateConversation, SendMessage, GetMessages
 
 app_name = "chat"
 
 
-urlpatterns = []
+urlpatterns = [
+    path("conversations/", CreateConversation.as_view()),
+    path("messages/", SendMessage.as_view()),
```
mehalyna, line 9: Without `name`, these routes cannot be referenced easily in Django templates or reverse lookups.
Add `name="route_name"` for all paths.

BackEnd/chat/urls.py
```diff
@@ -1,6 +1,15 @@
...
+    path(
+        "conversations/history/",
+        GetMessages.as_view(),
+        name="get_messages",
+    ),
```
mehalyna, line 14: "conversations/history/" does not allow fetching messages for a specific room. The API cannot fetch messages for a given conversation.
Modify the endpoint to include a room ID parameter.

BackEnd/chat/urls.py
mehalyna, line 9: "messages/" does not associate messages with a room in the URL. The request requires the room ID in the body, but URLs should be RESTful.
Modify the `SendMessage` endpoint to accept a room ID in the URL.

BackEnd/chat/views.py
mehalyna, line 1: `CreateConversation`, `SendMessage`, and `GetMessages` follow different naming patterns.
Inconsistent naming makes code harder to maintain.

Rename CreateConversation → ConversationCreateView
Rename SendMessage → MessageSendView
Rename GetMessages → MessageListView

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+                return Response(
+                    {
+                        "message": "You must be a participant in the conversation."
+                    },
+                    status=status.HTTP_403_FORBIDDEN,
+                )
+
+            existing_room = Room.objects.filter(
+                participant_ids=participant_ids
+            ).first()
```
mehalyna, line 35: The query `Room.objects.filter(participant_ids=participant_ids).first()` may not work as expected because MongoDB does not index lists for exact matches. The API may allow duplicate rooms if the `participant_ids` order is different.
Use `participant_ids__all=participant_ids` to correctly check if a room exists for the same participants.

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+
+            if existing_room:
+                return Response(
+                    {
+                        "message": "A room with this participant pair already exists."
+                    },
+                    status=status.HTTP_409_CONFLICT,
+                )
+            serializer.save()
```
mehalyna, line 44: The serializer does not explicitly pass participant_ids, so it may not store them correctly.
Pass `participant_ids` explicitly to `.save()`.

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
+            api_requested_sender_id = request.user.id
+        except AttributeError:
+            return Response(
+                {"message": "Authentication required."},
+                status=status.HTTP_401_UNAUTHORIZED,
+            )
+
+        serializer = MessageSerializer(data=request.data)
+        participant_ids = request.data["participant_ids"]
```
mehalyna, line 66: The `participant_ids` field does not exist in `MessageSerializer`, but the API tries to read it. Causes `KeyError` because `participant_ids` is not part of the request body.
Retrieve `room_id` instead and query the Room correctly.

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+        if serializer.is_valid():
+            sender_id = api_requested_sender_id
+
+            try:
+                room = Room.objects.get(participant_ids=participant_ids)
```
mehalyna, line 71: The line does not work in MongoDB, because `participant_ids` cannot be queried directly for an exact match.
Instead, query by `room_id`, which is a proper unique identifier.

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+                )
+
+            if sender_id not in room.participant_ids:
+                return Response(
+                    {
+                        "message": "Sender is not a participant of this conversation."
+                    },
+                    status=status.HTTP_403_FORBIDDEN,
+                )
+            serializer.save()
```
mehalyna, line 85: The `MessageSerializer` requires a room reference, but `.save()` does not pass it explicitly.
Pass `room=room` explicitly when saving the message.

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class GetMessages(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request, format=None):
+        try:
+            room = Room.objects.get(id=request.query_params.get("room_id"))
```
mehalyna, line 100: The API expects `room_id` from `request.query_params.get("room_id")`
RESTful best practices recommend using `room_id` in the URL path instead of query parameters.

Modify the URL in `urls.py`:
```
path("conversations/<str:room_id>/history/", GetMessages.as_view(), name="get_messages")
```
Retrieve `room_id` from `kwargs` instead:
```
room = Room.objects.get(id=kwargs["room_id"])
```

BackEnd/chat/views.py
```diff
@@ -0,0 +1,116 @@
...
+        except Room.DoesNotExist:
+            return Response(
+                {"message": "Room not found."},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        except Exception as e:
+            return Response(
+                {"message": f"An error occurred: {str(e)}"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
mehalyna, line 116: The API does not paginate messages, which can lead to performance issues if a room has thousands of messages.

mehalyna (changes requested):

BackEnd/chat/consumers.py
vladmorozov2: Thank you so much for your valuable feedback!
I will check in weekends

BackEnd/chat/views.py
vladmorozov2: fixed

BackEnd/chat/views.py
vladmorozov2: fixed

BackEnd/chat/views.py
vladmorozov2: fixed

BackEnd/chat/views.py
vladmorozov2: fuxed

BackEnd/chat/models.py
vladmorozov2: fuxed

BackEnd/chat/models.py
mehalyna: This is a good approach! Using MongoDB aggregation is indeed a more efficient and accurate way to check for duplicate rooms compared to a simple `.filter()` query, which doesn't properly handle lists in MongoDB indexes.
A potential improvement could be handling unordered participant lists by ensuring both the stored and incoming lists are sorted before comparison. Otherwise, two rooms with the same participants in a different order might not match correctly.

mehalyna (approved):

vladmorozov2 (commented):
nice

PR #102, opened 2025-02-28
Set up authorization functionality, which check the value of the role

IraMladanovych02:
- Define the tables needed to store user data/permissions/roles
- Check the JWT validation
- Provide a migration script 
- Verify a user’s permissions/roles from the database (e.g., checking if a user has admin role or certain permissions).


gateway/auth/jwt/jwt.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+    "errors"
+    "fmt"
+    "strings"
+
+    "github.com/golang-jwt/jwt/v4"
+)
+
+var jwtSecret = []byte("mysecret")
```
mehalyna, line 11: Use environment variables (`JWT_SECRET`) to load the secret key dynamically.
`var jwtSecret = []byte(os.Getenv("JWT_SECRET"))`

gateway/auth/jwt/jwt.go
```diff
@@ -0,0 +1,33 @@
...
+        if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
+            return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+        }
+        return jwtSecret, nil
+    })
+    if err != nil {
+        return nil, err
+    }
+    return token, nil
+}
```
mehalyna, line 31: The function only returns the token object, but does not extract claims.
Return `map[string]interface{}` along with `*jwt.Token` for easier claim extraction.

mehalyna, line 31: The function does not check if the token is expired. Extract `exp` (expiry) and validate against the current time.

gateway/auth/models/models.go
```diff
@@ -0,0 +1,25 @@
...
+    CreatedAt   time.Time
+}
+
+// Permission описує дозвол користувача
+type Permission struct {
+    ID          uint      `gorm:"primaryKey"`
+    Name        string    `gorm:"unique;not null"`
+    Description string
+    CreatedAt   time.Time
+}
```
mehalyna, line 19: The `Role` and `Permission` structs do not inherit from `gorm.Model`, which provides `CreatedAt`, `UpdatedAt`, and `DeletedAt` fields automatically. Use `gorm.Model` instead of manually adding `CreatedAt`.

mehalyna, line 19: Add `UpdatedAt` and `DeletedAt` fields. These fields are crucial for tracking updates and implementing soft deletes.

gateway/auth/models/models.go
```diff
@@ -0,0 +1,25 @@
...
+
+// RolePermission зв’язує роль з дозволом (багато-до-багатьох)
+type RolePermission struct {
+    RoleID       uint `gorm:"primaryKey"`
+    PermissionID uint `gorm:"primaryKey"`
+}
```
mehalyna, line 25: The struct lacks `gorm.ForeignKey` constraints for referential integrity. Use `gorm:"foreignKey"` to enforce relationships.

mehalyna (changes requested):

mehalyna (approved):

boghtml (approved):

PR #104, opened 2025-03-02
Notification service

isac-medison:
[initial for notifications service]
[add] feature send mails
[add] email templates
[add] docker for notifications service
Service uses the following virtual environment params

 EMAIL_HOST='smtp.gmail.com'
EMAIL_PORT="587"
EMAIL_USE_TLS=1
EMAIL_HOST_USER=""
EMAIL_HOST_PASSWORD=""
NOTIFICATIONS_HOST="0.0.0.0"
NOTIFICATIONS_PORT="8086"


BackEnd/notifications-service/src/main/scala/models/EnailRequest.scala
mehalyna, line 1: Looks like typo in file name `EnailRequest.scala`

BackEnd/notifications-service/src/main/scala/models/EnailRequest.scala
```diff
@@ -0,0 +1,9 @@
+package com.mycompany
+
+import spray.json._
+
+case class EmailRequest(email: String, link: String)
```
mehalyna, line 5: If `email` or `link` is empty, the application may crash or send invalid data.
Use require statements to enforce non-empty values. Provide a meaningful error message
```
require(email.nonEmpty, "Email cannot be empty.")
require(link.nonEmpty, "Link cannot be empty.")
```

docker-compose.notifications.yml
```diff
@@ -0,0 +1,21 @@
...
+services:
+  scala-mail-service:
+    image: sbtscala/scala-sbt:graalvm-ce-22.3.3-b1-java17_1.10.7_3.3.5
+    container_name: scala-mail-service
+    working_dir: /app
+    volumes:
+      - ./BackEnd/notifications-service:/app
+      - ./.env:/app/.env
+    env_file:
+      - .env
```
mehalyna, line 12: You are mounting `.env` to `/app/.env`, but Docker does not automatically load it inside the container. Scala and SBT will not recognize it. Explicitly load environment variables in command.

mehalyna, line 12: This should work without a volume mount

docker-compose.notifications.yml
```diff
@@ -0,0 +1,21 @@
...
+  scala-mail-service:
+    image: sbtscala/scala-sbt:graalvm-ce-22.3.3-b1-java17_1.10.7_3.3.5
+    container_name: scala-mail-service
+    working_dir: /app
+    volumes:
+      - ./BackEnd/notifications-service:/app
+      - ./.env:/app/.env
+    env_file:
+      - .env  
+    command: ["/bin/sh", "-c", "sbt reload && sbt compile && sbt run -deprecation "]
```
mehalyna, line 13: `sbt reload` → Might be unnecessary
`sbt run` → Could fail if dependencies are not fully compiled
Change command to:
```
    command: ["sbt", "clean", "compile", "run"]
```
This ensures:
- It removes old builds (`sbt clean`)
- Compiles fresh dependencies (`sbt compile`)
- Runs the app properly (`sbt run`)

docker-compose.notifications.yml
mehalyna, line 1: Currently, there's no healthcheck, meaning Docker won’t know if the service is healthy.
Add a healthcheck to verify if the service is running:
```
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:${NOTIFICATIONS_PORT}/health"]
      interval: 10s
      timeout: 5s
      retries: 3
```
Your Scala Akka HTTP app should expose a `/health` endpoint.

mehalyna (changes requested):

BackEnd/notifications-service/src/main/scala/constants/EmailConstants.scala
```diff
@@ -0,0 +1,5 @@
+package com.mycompany  
+object EmailConstants {
```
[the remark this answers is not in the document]
v2496s: Using a sealed trait makes constants unreachable in separate files. I would also like to use some type for my const.
Trait does not allow params, while sealed does not allow use from different files.

v2496s:
I have decided to move   
  val emailHost: String = EmailConstants.envVars("EMAIL_HOST")
  val emailPort: String = EmailConstants.envVars("EMAIL_PORT")
  val emailHostUser: String = EmailConstants.envVars("EMAIL_HOST_USER")
  val emailHostPassword: String = EmailConstants.envVars("EMAIL_HOST_PASSWORD")
  to email sender, as it will be created once and no need to pass sensitive params to some methods, it will be in common with all email endpoints
  Same works for extraction methods, they are now being placed on constants file

mehalyna (approved):

boghtml (approved):

PR #120, opened 2025-03-04
Basic setup for authentication service

sofiahnatiuk:
1. Set up basic structure for the authentication service
2. Added Dockerfile, requirements.txt, .gitignore for the authentication service
3. Copied authentication logic from the authentication app of the main project


authentication_service/authentication/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ENV HOME=/home/authentication
+ENV APP_HOME=$HOME/app
+
+WORKDIR $APP_HOME
+
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PYTHONUNBUFFERED=1
+
+RUN apk add --no-cache gcc python3-dev musl-dev
```
mehalyna: `gcc`, `python3-dev`, and `musl-dev` are build dependencies that are usually only needed for compiling packages. After installing dependencies, remove them to reduce image size.
Use a multi-stage build or remove build dependencies after install
```
RUN apk add --no-cache --virtual .build-deps gcc python3-dev musl-dev \
    && pip install --upgrade pip && pip install --no-cache-dir -r requirements.txt \
    && apk del .build-deps
```

authentication_service/authentication/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY ./requirements.txt .
+
+RUN pip install --upgrade pip && pip install --no-cache-dir -r requirements.txt
```
mehalyna: The `pip install` runs before switching to USER authentication, meaning it's installed as root. This may cause permission issues when running inside a container.

authentication_service/authentication/Dockerfile
mehalyna, line 1: Multiple `RUN` statements increase layer size. Combine them to reduce the number of layers and speed up builds.
```
RUN apk add --no-cache --virtual .build-deps gcc python3-dev musl-dev \
    && pip install --upgrade pip && pip install --no-cache-dir -r requirements.txt \
    && apk del .build-deps \
    && rm -rf /var/cache/apk/*
```

authentication_service/authentication/authentication/settings.py
```diff
@@ -0,0 +1,283 @@
...
+    "http://127.0.0.1",
+    "http://0.0.0.0",
+]
+
+CORS_ALLOW_HEADERS = list(default_headers) + [
+    "Access-Control-Expose-Headers",
+    "Access-Control-Allow-Headers",
+    "Access-Control-Allow-Origin",
+    "Content-Type",
+]
```
mehalyna: `CORS_ALLOW_ALL_ORIGINS = False` contradicts `CORS_ORIGIN_WHITELIST`, which allows external domains.
Use only `CORS_ALLOWED_ORIGINS`

authentication_service/authentication/authentication/settings.py
```diff
@@ -0,0 +1,283 @@
...
+
+# Database
+# https://docs.djangoproject.com/en/4.2/ref/settings/#databases
+
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": BASE_DIR / "db.sqlite3",
+    }
+}
```
mehalyna: SQLite is good for development, but not for production. Consider using PostgreSQL or MySQL in production.

authentication_service/authentication/authentication/settings.py
```diff
@@ -0,0 +1,283 @@
...
+TOKEN_EXPIRATION_TIME = timedelta(days=14)
+
+# SMTP
+EMAIL_BACKEND = config("EMAIL_BACKEND")
+EMAIL_HOST = config("EMAIL_HOST")
+EMAIL_PORT = config("EMAIL_PORT")
+EMAIL_USE_TLS = config("EMAIL_USE_TLS")
+EMAIL_HOST_USER = config("EMAIL_HOST_USER")
+EMAIL_HOST_PASSWORD = config("EMAIL_HOST_PASSWORD")
```
mehalyna: If any email-related environment variable is missing, sending emails may silently fail. Set default values

authentication_service/authentication/authentication/settings.py
```diff
@@ -0,0 +1,283 @@
...
+            "level": "DEBUG",
+            "propagate": False,
+        },
+    },
+}
+# ReCaptcha V2 Invisible
+RECAPTCHA_V2_PRIVATE_KEY = config("RECAPTCHA_V2_PRIVATE_KEY")
+RECAPTCHA_URL = config("RECAPTCHA_URL")
+
+DJANGO_SETTINGS_MODULE = config("DJANGO_SETTINGS_MODULE")
```
mehalyna, line 284: This should be set in the system environment instead. Remove this from `settings.py`. Instead, set it before running Django:
```
export DJANGO_SETTINGS_MODULE=authentication.settings
```

authentication_service/authentication/custom_auth/authentication.py
```diff
@@ -0,0 +1,35 @@
...
+
+            if not user.is_active:
+                raise AuthenticationFailed("User account is deactivated.")
+
+            return user, validated_token
+
+        except (InvalidToken, TokenError) as e:
+            raise AuthenticationFailed(f"Invalid or expired token: {str(e)}")
+        except AuthenticationFailed as e:
+            raise AuthenticationFailed(f"Authentication failed. Please log in again: {str(e)}")
```
mehalyna, line 35: The `AuthenticationFailed` exception is already raised by `get_user(validated_token)` if the user is invalid. The extra `except AuthenticationFailed as e` adds redundancy and is not needed.

authentication_service/authentication/custom_auth/authentication.py
```diff
@@ -0,0 +1,35 @@
...
+        except Exception as e:
+            raise AuthenticationFailed(f"An error occurred during authentication: {str(e)}")
```
mehalyna: If `get_validated_token(raw_token`) fails first, the user object is never retrieved. This may lead to a misleading error message in the Exception block.

authentication_service/authentication/custom_auth/factories.py
```diff
@@ -0,0 +1,16 @@
...
+
+class UserFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = CustomUser
+        django_get_or_create = ("email",)
+
+    email = factory.Sequence(lambda n: f"test{n + 1}@test.com")
+    name = factory.Sequence(lambda n: f"Test person {n + 1}")
+    surname = factory.Sequence(lambda n: f"Test person {n + 1} surname")
+    password = ""
```
mehalyna: Django requires hashed passwords. If you create a user like this, they will not be able to log in.
Use `set_password()` to hash the password

authentication_service/authentication/custom_auth/factories.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class UserFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = CustomUser
+        django_get_or_create = ("email",)
+
+    email = factory.Sequence(lambda n: f"test{n + 1}@test.com")
+    name = factory.Sequence(lambda n: f"Test person {n + 1}")
+    surname = factory.Sequence(lambda n: f"Test person {n + 1} surname")
```
mehalyna: This does not create distinct test users. Differentiate the name and surname
```
name = factory.Faker("first_name")
surname = factory.Faker("last_name")
```

mehalyna: `factory.Sequence` is predictable, but `Faker` generates more natural test data.

authentication_service/authentication/custom_auth/managers.py
```diff
@@ -0,0 +1,43 @@
...
+        :param password (str): The user's password.
+        :param **extra_fields: Additional user fields.
+
+        Returns:
+            CustomUser: The created user instance.
+        """
+
+        email = self.normalize_email(email)
+        user = self.model(email=email, **extra_fields)
+        user.set_password(password)
```
mehalyna, line 26: `set_password()` allows `None` as a password, which means a user could be created without a password.
This can cause security issues where users can exist without valid authentication.

authentication_service/authentication/custom_auth/managers.py
```diff
@@ -0,0 +1,43 @@
...
+        
+        :param email (str): The email address of the superuser.
+        :param password (str): The superuser's password.
+        :param **extra_fields: Additional superuser fields.
+
+        Returns:
+            CustomUser: The created superuser instance.
+        """
+        extra_fields.setdefault('is_staff', True)
+        extra_fields.setdefault('is_active', True)
```
mehalyna: The line is redundant, because you later set:
```
user.is_active = True
```

authentication_service/authentication/custom_auth/models.py
```diff
@@ -0,0 +1,76 @@
...
+        Args:
+            perm (str): The permission codename to check.
+            obj (optional): The object to check the permission against. Defaults to None.
+
+        Returns:
+            bool: True if the user has the permission, False otherwise.
+        """
+        if self.is_superuser:
+            return True
+        return self.roles.filter(permissions__codename=perm).exists()
```
mehalyna: The method filters through roles and permissions every time it's called, which could lead to performance issues if used frequently. Instead of querying the database each time, cache the result or use Django's built-in permission system.

mehalyna: If a user has many roles, querying `self.roles.filter(permissions__codename=perm).exists()` may be slow. Consider prefetching related permissions in queries. Use `prefetch_related()` in queries
```
def get_all_permissions(self):
    return set(self.roles.prefetch_related("permissions").values_list("permissions__codename", flat=True))
```

authentication_service/authentication/custom_auth/models.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Fields:
+        name (CharField): The human-readable name of the permission.
+        codename (CharField): The unique codename of the permission.
+    """
+    name = models.CharField(max_length=100, unique=True)
+    codename = models.CharField(max_length=100, unique=True)
+
+    def __str__(self):
+        return self.name
```
mehalyna: Django already has a `Permission` model under `django.contrib.auth.models.Permission`, which is widely used and well-integrated. Custom permissions may break compatibility with Django’s permission framework.
Consider extending Django’s permission model

authentication_service/authentication/custom_auth/serializers.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def validate(self, attrs):
+        user = self.context["request"].user
+        if user.check_password(attrs["new_password"]):
+            raise serializers.ValidationError(
+                {"new_password": "New password must be different from the current password."})
+        if attrs["new_password"] != attrs["confirm_password"]:
+            raise serializers.ValidationError({"confirm_password": "Passwords do not match."})
+        return attrs
+
```
mehalyna: Django expects lists for field-specific errors, but `{"new_password": errors}` creates a nested dictionary. This can break error handling in DRF.

authentication_service/authentication/custom_auth/serializers.py
```diff
@@ -0,0 +1,299 @@
...
+        if user.check_password(new_password):
+            custom_errors["new_password"].append("New password must be different from the current password.")
+
+        if custom_errors:
+            raise serializers.ValidationError(custom_errors)
+
+        if not default_token_generator.check_token(user, data["token"]):
+            raise serializers.ValidationError({"token": "Token is invalid or expired"})
+
+        return {"user": user, "new_password": new_password}
```
mehalyna: The validation checks if the new password is different, but it doesn't enforce Django's built-in password policies.

authentication_service/authentication/custom_auth/serializers.py
```diff
@@ -0,0 +1,299 @@
...
+        return data
+
+
+class LogoutSerializer(serializers.Serializer):
+    refresh = serializers.CharField(required=True)
+
+    def validate(self, data):
+        request = self.context.get('request')
+        auth_header = request.headers.get('Authorization', '')
+        access_token = auth_header.split()[1]
```
mehalyna, line 151: If Authorization is missing, `auth_header.split()[1]` raises an `IndexError`.

authentication_service/authentication/custom_auth/serializers.py
```diff
@@ -0,0 +1,299 @@
...
+            decoded_uid_bytes = urlsafe_base64_decode(data["uid"])
+            decoded_uid = force_str(decoded_uid_bytes)
+        except Exception:
+            raise serializers.ValidationError({"uid": "Invalid uid provided."})
+
+        try:
+            user = User.objects.get(pk=decoded_uid)
+        except (User.DoesNotExist, ValueError, TypeError):
+            raise serializers.ValidationError(
+                {"uid": "User does not exist. Please, try again."}
```
mehalyna, line 243: If decoding fails, it raises a generic `ValidationError` without specifying the issue.

authentication_service/authentication/custom_auth/serializers.py
```diff
@@ -0,0 +1,299 @@
...
+        company_data = validated_data.pop("company")
+        user = User.objects.create(
+            email=validated_data["email"],
+            name=validated_data["name"],
+            surname=validated_data["surname"],
+        )
+        user.set_password(validated_data["password"])
+        logger.info(f"Saving user {user.email}")
+        user.save()
+        return user
```
mehalyna: The `User` model is saved twice, once explicitly in `create()` and once by `set_password()`.
Use `update_fields` when setting password
```
user.set_password(validated_data["password"])
user.save(update_fields=["password"])
```

authentication_service/authentication/custom_auth/tests/test_password_reset.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_password_reset_confirm_invalid_token(self):
+        response = self.client.post(
+            "/api/auth/password-reset-confirm/",
+            {"token": "invalidtoken", "new_password": "NewSecurePass123!"},
+            format="json"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("Token is invalid or expired.", response.data["error"])
```
mehalyna: The test expects "Token is invalid or expired.", but the API might return a different error message (e.g., "Invalid signature"). If Django’s `TimestampSigner` raises a different exception, the test may fail unexpectedly.
Allow flexible error messages
```
self.assertIn("invalid", response.data["error"].lower())  # Accepts variations like "Invalid signature"
```

authentication_service/authentication/custom_auth/tests/test_password_reset.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_password_reset_confirm_weak_password(self):
+        token = signer.sign(f"{self.user.pk}:{self.user.password}")
+        response = self.client.post(
+            "/api/auth/password-reset-confirm/",
+            {"token": token, "new_password": "123"},
+            format="json"
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("security requirements", response.data.get("error", "").lower())
```
mehalyna: The test expects a generic "security requirements" error, but Django’s built-in validate_password() may return:
"This password is too short."
"This password is too common."
"This password is entirely numeric."
Match error more precisely
```
self.assertTrue(
    any(msg in response.data.get("error", "").lower() for msg in ["too short", "too common", "numeric"])
)
```

authentication_service/authentication/custom_auth/tests/test_password_reset.py
```diff
@@ -0,0 +1,109 @@
...
+            "/api/auth/password-reset-confirm/",
+            {"token": "invalidtoken", "new_password": "NewSecurePass123!"},
+            format="json"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("Token is invalid or expired.", response.data["error"])
+
+    def test_password_reset_confirm_weak_password(self):
+        token = signer.sign(f"{self.user.pk}:{self.user.password}")
```
mehalyna, line 83: If the user’s password changes before token verification, the token will be invalid. `signer.sign(f"{self.user.pk}:{self.user.password}")` includes `self.user.password`, which changes when the password is updated.
Use a stable identifier (e.g., User ID)
```
token = signer.sign(f"{self.user.pk}")  # Remove dependency on password hash
```

authentication_service/authentication/custom_auth/tests/test_user_activation.py
```diff
@@ -0,0 +1,71 @@
...
+        expired_token = self.signer.sign(str(self.test_user.pk))
+        expired_time = timezone.now() - timedelta(hours=2)
+
+        with patch('django.core.signing.TimestampSigner.unsign') as mock_unsign:
+            mock_unsign.side_effect = BadSignature
+
+            response = self.client.get(f"{url}?token={expired_token}")
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), {"error": "Invalid or expired token."})
```
mehalyna: The test patches `unsign()` to raise `BadSignature`, but this does not truly simulate an expired token. `BadSignature` is raised for invalid tokens, not necessarily for expired ones.
Simulate expiry by setting an older Timestamp
```
with patch('django.core.signing.TimestampSigner.unsign') as mock_unsign:
    mock_unsign.side_effect = lambda token, max_age: BadSignature if max_age < 3600 else str(self.test_user.pk)
    response = self.client.get(f"{url}?token={expired_token}")
```

authentication_service/authentication/custom_auth/tests/test_user_activation.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_account_activation_invalid_token(self):
+        invalid_token = "invalid_token"
+        response = self.client.get(f"{url}?token={invalid_token}")
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), {"error": "Invalid or expired token."})
+
+    def test_account_activation_user_not_found(self):
+        invalid_token = self.signer.sign("999")
```
mehalyna: If a user with `pk=999` actually exists in the test database, this test may fail unpredictably

authentication_service/authentication/custom_auth/tests/test_user_activation.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertEqual(response.json(), {"detail": "Account activated successfully."})
+
+    def test_account_activation_missing_token(self):
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), {"detail": "No token provided."})
+
+    def test_account_activation_invalid_token(self):
+        invalid_token = "invalid_token"
```
mehalyna, line 40: If a random string like "invalid_token" is used, `BadSignature` is raised before even reaching the expiration check.
This may cause the error message to mislead users.

authentication_service/authentication/custom_auth/tests/test_user_activation.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+url = reverse('authentication:activate')
+
+class AccountActivationAPITests(APITestCase):
+    def setUp(self):
+        patcher = patch(
+            "authentication.serializers.verify_recaptcha", return_value=True
+        )
+        self.mock_verify_recaptcha = patcher.start()
```
mehalyna: The test patches `verify_recaptcha` but never actually calls it. If the API does not require reCAPTCHA, this is an unnecessary mock.

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
```diff
@@ -0,0 +1,85 @@
...
+        self.test_user_token = self.client.post(
+            path="/api/auth/token/login/",
+            data={
+                "email": "test@test.com",
+                "password": "Test1234",
+                "captcha": "dummy_captcha",
+            },
+        ).data["auth_token"]
+        token = Token.objects.get(key=self.test_user_token)
+        token.created -= timedelta(days=15)
```
mehalyna: Django’s `Token` model might not recognize direct updates to `token.created` since it may not be included in the `save()` method. The test could fail because the token expiration logic may still use the original creation time.
Explicitly set and save created field
```
Token.objects.filter(key=self.test_user_token).update(created=token.created - timedelta(days=15))
```

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
```diff
@@ -0,0 +1,85 @@
...
+        self.user.save()
+
+        self.test_user_token = self.client.post(
+            path="/api/auth/token/login/",
+            data={
+                "email": "test@test.com",
+                "password": "Test1234",
+                "captcha": "dummy_captcha",
+            },
+        ).data["auth_token"]
```
mehalyna: If login fails, `self.test_user_token = response.data["auth_token"]` will raise a `KeyError`. The test will crash instead of failing gracefully. Handle missing `auth_token`

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
```diff
@@ -0,0 +1,85 @@
...
+        token = Token.objects.get(key=self.test_user_token)
+        token.created -= timedelta(days=10)
```
mehalyna: comment line 42

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
```diff
@@ -0,0 +1,85 @@
...
+        self.user.save()
+
+        self.test_user_token = self.client.post(
+            path="/api/auth/token/login/",
+            data={
+                "email": "test@test.com",
+                "password": "Test1234",
+                "captcha": "dummy_captcha",
+            },
+        ).data["auth_token"]
```
mehalyna, line 73: comment lines 33-40

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        response = self.client.get(path="/api/auth/users/me/")
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertEqual(
+            {"detail": "Your session has expired. Please login again."},
+            response.json(),
+        )
+
+    def test_user_autologout_after_10_days(self):
+        self.user.set_password("Test1234")
```
mehalyna, line 63: `self.user.set_password("Test1234")` hashes the password but does not save the user. The user might still have an empty password, leading to login failures.

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
```diff
@@ -0,0 +1,85 @@
...
+        self.user = UserFactory(
+            email="test@test.com", name="Test", surname="Test"
+        )
+        self.profile = ProfileCompanyFactory.create(
+            person=self.user,
+            official_name="Test Official Startup",
+        )
+
+    def test_user_autologout_after_14_days(self):
+        self.user.set_password("Test1234")
```
mehalyna, line 32: `self.user.set_password("Test1234")` hashes the password but does not save the user. The user might still have an empty password, leading to login failures.

authentication_service/authentication/custom_auth/tests/test_user_factory.py
```diff
@@ -0,0 +1,12 @@
+from django.test import TestCase
+
+from custom_auth.factories import UserFactory
+
+
+class TestFactories(TestCase):
+    def test_user_factory(self):
+        user = UserFactory()
+        self.assertIsNotNone(user.email)
```
mehalyna, line 12: `assertIsNotNone(user.email)` only checks if the field exists, but the email could still be an empty string.
The test might pass even if `user.email = ""`, which would not be a valid user.

authentication_service/authentication/custom_auth/validate_recaptcha.py
```diff
@@ -0,0 +1,19 @@
...
+    Validates the reCAPTCHA token with Google's API.
+    """
+    recaptcha_url = settings.RECAPTCHA_URL
+    recaptcha_data = {
+        "secret": settings.RECAPTCHA_V2_PRIVATE_KEY,
+        "response": token,
+    }
+    response = requests.post(recaptcha_url, data=recaptcha_data)
+    result = response.json()
+    logger.info("Successful reCAPTCHA")
```
mehalyna: The logger always logs "Successful reCAPTCHA", even if `result.get("success", False)` is `False`. This misleads developers into thinking the validation was successful when it actually failed.

authentication_service/authentication/custom_auth/validate_recaptcha.py
```diff
@@ -0,0 +1,19 @@
...
+    return result.get("success", False)
```
mehalyna: If Google’s API is unreachable, the function will raise a `requests.exceptions.RequestException` instead of returning `False`. This can cause unhandled crashes in your application.

authentication_service/authentication/custom_auth/validate_recaptcha.py
```diff
@@ -0,0 +1,19 @@
...
+def verify_recaptcha(token):
+    """
+    Validates the reCAPTCHA token with Google's API.
+    """
+    recaptcha_url = settings.RECAPTCHA_URL
+    recaptcha_data = {
+        "secret": settings.RECAPTCHA_V2_PRIVATE_KEY,
+        "response": token,
+    }
+    response = requests.post(recaptcha_url, data=recaptcha_data)
```
mehalyna: If Google’s API takes too long to respond, the request might hang indefinitely. This can cause performance issues.
Set a Timeout
```
response = requests.post(recaptcha_url, data=recaptcha_data, timeout=5)
```

authentication_service/authentication/custom_auth/validate_recaptcha.py
```diff
@@ -0,0 +1,19 @@
...
+
+def verify_recaptcha(token):
+    """
+    Validates the reCAPTCHA token with Google's API.
+    """
+    recaptcha_url = settings.RECAPTCHA_URL
+    recaptcha_data = {
+        "secret": settings.RECAPTCHA_V2_PRIVATE_KEY,
+        "response": token,
+    }
```
mehalyna: If `settings.RECAPTCHA_V2_PRIVATE_KEY` is missing or `None`, the request will still be sent with an invalid key. Google may block or ignore your reCAPTCHA requests.

authentication_service/authentication/custom_auth/validate_password.py
```diff
@@ -0,0 +1,69 @@
...
+    Raises ValidationError if the password is weak.
+    """
+    if not re.search(r'[!@#$%^&*(),.?":{}|<>]', password_value):
+        raise ValidationError(
+            "Password must include at least one special character (e.g., !@#$%^&*)."
+        )
+
+    common_passwords = [
+        "123456", "password", "qwerty", "admin", "letmein", "12345678", "123123", "111111"
+    ]
```
mehalyna, line 46: The list of common passwords does not update. Attackers use updated leaked password lists, making this less effective over time.
Use haveibeenpwned.com API, or https://gist.github.com/roycewilliams/226886fd01572964e1431ac8afc999ce, or expand the List

authentication_service/authentication/custom_auth/validate_password.py
```diff
@@ -0,0 +1,69 @@
...
+    if password_value.lower() in common_passwords:
+        raise ValidationError(
+            "Password is too common or easily guessable."
+        )
+
+    sequences = [
+        "1234", "2345", "3456", "4567", "5678", "6789", "7890",
+        "abcd", "bcde", "cdef", "defg", "efgh", "fghi", "ghij",
+        "qwert", "werty", "ertyui", "rtyui", "tyuio", "yuiop",
+    ]
```
mehalyna, line 57: The function checks for "abcd" and "1234", but misses patterns like "zxcv" and "0987". Attackers use keyboard-based dictionary attacks.
Add more keyboard patterns

authentication_service/authentication/custom_auth/urls.py
```diff
@@ -0,0 +1,40 @@
...
+    AccountActivationView,
+)
+
+app_name = "authentication"
+
+urlpatterns = [
+    path('auth/register/', UserRegistrationView.as_view(), name='register'),
+    path('auth/activate/', AccountActivationView.as_view(), name='activate'),
+    path('auth/logout/', LogoutView.as_view(), name='logout'),
+    path('auth/login/', TokenObtainPairView.as_view(), name='login'),
```
mehalyna: The login endpoint  is using `TokenObtainPairView`, but a custom version (`CustomTokenObtainPairView`) is also defined.
If `CustomTokenObtainPairView` adds extra claims (e.g., `email`, `is_staff`), these will not be included when users log in via `/auth/login/`.
Replace `TokenObtainPairView` with `CustomTokenObtainPairView`

authentication_service/authentication/custom_auth/urls.py
```diff
@@ -0,0 +1,40 @@
...
+    ),
+    path(
+        "auth/password-reset-confirm/",
+        PasswordResetConfirmView.as_view(),
+        name="password_reset_confirm"
+    ),
+    path("auth/password-change/", PasswordChangeView.as_view(), name="password_change"),
+
+    # JWT implementation
+    path('auth/jwt/create/', CustomTokenObtainPairView.as_view(), name='jwt_create'),
```
mehalyna: Both:
```
path('auth/login/', CustomTokenObtainPairView.as_view(), name='login'),
path('auth/jwt/create/', CustomTokenObtainPairView.as_view(), name='jwt_create'),
```
Point to the same view (`CustomTokenObtainPairView`).
This creates duplicate endpoints for obtaining a JWT token.
Remove one of these for cleaner API design
```
# Remove jwt_create if login does the same thing
path('auth/jwt/create/', CustomTokenObtainPairView.as_view(), name='jwt_create'),
```
OR
```
# Rename /auth/login/ to /auth/jwt/create/ if JWT naming is preferred
path('auth/jwt/create/', CustomTokenObtainPairView.as_view(), name='login'),
```

authentication_service/authentication/custom_auth/urls.py
```diff
@@ -0,0 +1,40 @@
+from django.urls import include, path, re_path
```
mehalyna: re_path is imported but never used.

authentication_service/authentication/custom_auth/urls.py
mehalyna, line 1: Django REST Framework by default requires a trailing slash (`/auth/jwt/create/`), but many API clients expect no trailing slash. This could lead to inconsistent API behavior when using clients like Postman or mobile apps.
Explicitly set `trailing_slash=False` in JWT Paths
```
from rest_framework.routers import SimpleRouter

router = SimpleRouter(trailing_slash=False)

urlpatterns += [
    path('auth/jwt/create', CustomTokenObtainPairView.as_view(), name='jwt_create'),
    path('auth/jwt/refresh', TokenRefreshView.as_view(), name='jwt_refresh'),
    path('auth/jwt/verify', TokenVerifyView.as_view(), name='jwt_verify'),
]
```

authentication_service/authentication/custom_auth/tests/test_user_self.py
mehalyna, line 1: Consider all the suggestions via comments in previous test cases, pls

authentication_service/authentication/custom_auth/tests/test_user_registration.py
mehalyna, line 1: Consider all the suggestions via comments in previous test cases, pls

authentication_service/authentication/custom_auth/tests/test_user_logout.py
mehalyna, line 1: Consider all the suggestions via comments in previous test cases, pls

authentication_service/authentication/custom_auth/tests/test_user_login.py
mehalyna, line 1: Consider all the suggestions via comments in previous test cases, pls

mehalyna (changes requested):

authentication_service/authentication/Dockerfile
sofiahnatiuk: Updated Dockerfile to prevent permission issues inside a container

authentication_service/authentication/authentication/settings.py
sofiahnatiuk: Updated CORS parameters in settings.py

authentication_service/authentication/authentication/settings.py
sofiahnatiuk: DJANGO_SETTINGS_MODULE is set in system environment

authentication_service/authentication/custom_auth/authentication.py
sofiahnatiuk: Updated the file to check for token errors before accessing user object

authentication_service/authentication/custom_auth/managers.py
sofiahnatiuk: Added a check for empty password

authentication_service/authentication/custom_auth/models.py
sofiahnatiuk: Removed Permission model and has_permission method from CustomUser model, because this logic will be handled at the API Gateway in the future

authentication_service/authentication/custom_auth/serializers.py
sofiahnatiuk: New password is validated with validate_password_long(), validate_password_include_symbols(), validate_password_strength()

authentication_service/authentication/custom_auth/serializers.py
sofiahnatiuk: Added a check for missing auth_header

authentication_service/authentication/custom_auth/serializers.py
sofiahnatiuk: Added more custom errors

authentication_service/authentication/custom_auth/tests/test_password_reset.py
sofiahnatiuk: If the user's password changes before token verification, the token will be invalid, which prevents users from reusing old password reset links

authentication_service/authentication/custom_auth/tests/test_user_activation.py
sofiahnatiuk: Expired tokens are tested in `test_account_activation_expired_token`

NikitaPash (approved):

authentication_service/authentication/custom_auth/tests/test_user_autologout.py
sofiahnatiuk: The user is saved on line 64 with `self.user.save()`

authentication_service/authentication/custom_auth/urls.py
sofiahnatiuk: Removed /jwt/create/

authentication_service/authentication/custom_auth/tests/test_user_self.py
sofiahnatiuk: Will update this test file in a future pull request, because there will be changes in the CustomUser model

authentication_service/authentication/custom_auth/validate_password.py
sofiahnatiuk: Will create a separate task to improve validate_password logic and will update this file in a future pull request

mehalyna (approved):
Well done!

PR #135, opened 2025-03-07
31 notification

vladmorozov2:
Implemented notifications


BackEnd/chat_notifications/consumer.py
```diff
@@ -0,0 +1,72 @@
...
+        if not self.scope["user"].is_authenticated:
+            logger.warning("Unauthenticated user attempted to connect.")
+            await self.close()
+            return
+
+        self.room_group_name = f'notifications_{self.scope["user"].id}'
+
+        if self.channel_layer is None:
+            logger.error("channel_layer is not configured properly.")
+            await self.close()
```
mehalyna: If `channel_layer` is `None`, the consumer closes the connection immediately. This prevents any debugging insight into why `channel_layer` failed. Raise an exception with a detailed message
```
if self.channel_layer is None:
    logger.error("channel_layer is not configured. Ensure Django Channels is set up with a proper backend (e.g., Redis).")
    await self.close()
    return
```

BackEnd/chat_notifications/consumer.py
```diff
@@ -0,0 +1,72 @@
...
+                await self.channel_layer.group_discard(
+                    self.room_group_name, self.channel_name
+                )
+                logger.info(
+                    f"User {self.scope['user'].id} disconnected from notifications."
+                )
+            except KeyError as e:
+                logger.error(f"KeyError in group_discard: {str(e)}")
+            except RuntimeError as e:
+                logger.error(f"Runtime error in group_discard: {str(e)}")
```
mehalyna: If `connect()` fails early (e.g., due to `channel_layer` issues), `disconnect()` may still try to call `group_discard` on an undefined `room_group_name`. This could cause unexpected errors on disconnect.
Ensure `room_group_name` exists before `group_discard`
```
if hasattr(self, "room_group_name") and self.channel_layer is not None:
    try:
        await self.channel_layer.group_discard(self.room_group_name, self.channel_name)
        logger.info(f"User {self.scope['user'].id} disconnected from notifications.")
    except Exception as e:
        logger.error(f"Error during group_discard: {str(e)}")
```

BackEnd/chat_notifications/consumer.py
```diff
@@ -0,0 +1,72 @@
...
+        except json.JSONDecodeError as e:
+            logger.error(
+                f"JSON encoding error while sending notification: {str(e)}"
+            )
+        except ChannelFull as e:
+            logger.error(f"Channel layer is full, message dropped: {str(e)}")
+        except ConnectionError as e:
+            logger.error(
+                f"Connection error while sending notification: {str(e)}"
+            )
```
mehalyna: If `channel_layer` is full, the notification is simply dropped. This could lead to missed notifications without attempting a retry. Implement a basic retry mechanism
```
import asyncio

async def send_notification(self, event):
    """Send notification to the WebSocket client."""
    try:
        notification = event.get("notification", "No notification data provided")
        await self.send(text_data=json.dumps({"notification": notification}))

    except json.JSONDecodeError as e:
        logger.error(f"JSON encoding error while sending notification: {str(e)}")

    except ChannelFull as e:
        logger.warning(f"Channel layer is full, retrying in 1 second: {str(e)}")
        await asyncio.sleep(1)
        try:
            await self.send(text_data=json.dumps({"notification": notification}))
        except ChannelFull as e:
            logger.error(f"Channel layer is still full, message dropped: {str(e)}")

    except ConnectionError as e:
        logger.error(f"Connection error while sending notification: {str(e)}")
```

BackEnd/chat_notifications/models.py
```diff
@@ -0,0 +1,26 @@
...
+from django.utils import timezone
+from chat.models import Message, Room
+
+
+class Notification(Document):
+    recipient_id = IntField(required=True)
+    message = ReferenceField(Message, required=True, reverse_delete_rule=2)
+    is_read = BooleanField(default=False)
+    room = ReferenceField(Room, required=True, reverse_delete_rule=2)
+    created_at = DateTimeField(default=lambda: timezone.now())
```
mehalyna: Using `lambda: timezone.now()` creates a new function instance each time the model is evaluated. This is unnecessary because `default=timezone.now` already calls the function dynamically. Use `default=timezone.now` directly

BackEnd/chat_notifications/models.py
```diff
@@ -0,0 +1,26 @@
...
+)
+from django.utils import timezone
+from chat.models import Message, Room
+
+
+class Notification(Document):
+    recipient_id = IntField(required=True)
+    message = ReferenceField(Message, required=True, reverse_delete_rule=2)
+    is_read = BooleanField(default=False)
+    room = ReferenceField(Room, required=True, reverse_delete_rule=2)
```
mehalyna: The code uses magic numbers (2) for `reverse_delete`_rule. This makes the code hard to read and maintain. Use `mongoengine.CASCADE` instead of 2

BackEnd/chat_notifications/signals.py
mehalyna, line 1: The signal assumes `Notification.objects.create(...)` works like Django ORM, but MongoEngine does not use  `objects.create(...)`. This could cause a `TypeError` or `AttributeError` if the `Notification` model is using `MongoEngine`.
Use `.save()` instead for MongoEngine
```
from chat_notifications.models import Notification

for recipient_id in recipient_ids:
    notification = Notification(
        recipient_id=recipient_id,
        message=message,
        room=room,
    )
    notification.save()
```

BackEnd/chat_notifications/signals.py
```diff
@@ -0,0 +1,33 @@
...
+
+            # Send the notification to the channel layer (WebSocket)
+            channel_layer = get_channel_layer()
+            async_to_sync(channel_layer.group_send)(
+                f"notifications_{recipient_id}",  # Unique group for each user
+                {
+                    "type": "send_notification",
+                    "notification": f"New message from {message.sender_id}",
+                },
+            )
```
mehalyna: The channel layer is retrieved in every loop iteration, which creates unnecessary overhead. This reduces performance, especially when many notifications are sent at once.
Retrieve `channel_layer` nce
```
channel_layer = get_channel_layer()
for recipient_id in recipient_ids:
    notification = Notification(
        recipient_id=recipient_id,
        message=message,
        room=room,
    )
    notification.save()

    async_to_sync(channel_layer.group_send)(
        f"notifications_{recipient_id}",
        {
            "type": "send_notification",
            "notification": f"New message from {message.sender_id}",
        },
    )
```

mehalyna: The WebSocket notification message is a string, but most clients expect structured JSON.
Send JSON in WebSocket message
```
import json

async_to_sync(channel_layer.group_send)(
    f"notifications_{recipient_id}",
    {
        "type": "send_notification",
        "notification": json.dumps({
            "message": f"New message from {message.sender_id}",
            "room_id": str(room.id),
            "recipient_id": recipient_id,
        }),
    },
)
```

BackEnd/chat_notifications/signals.py
```diff
@@ -0,0 +1,33 @@
...
+from channels.layers import get_channel_layer
+from asgiref.sync import async_to_sync
+
+
+def create_notification(
+    sender, message, room, recipient_ids, created, **kwargs
+):
+
+    if created:
+        from chat_notifications.models import Notification
```
mehalyna, line 11: The Notification model is imported inside the function, even if `created` is False. This slows down execution for updates that don’t require a notification.

mehalyna (changes requested):

BackEnd/chat_notifications/signals.py
vladmorozov2: I fixed, but it works what a problem

BackEnd/chat_notifications/consumer.py
vladmorozov2: fixed

BackEnd/chat_notifications/signals.py
vladmorozov2: I could not put this import out of function it gives me circular import error

BackEnd/chat_notifications/signals.py
vladmorozov2: fixed

vladmorozov2: fixed

BackEnd/chat_notifications/models.py
vladmorozov2: fixed

BackEnd/chat_notifications/models.py
vladmorozov2: fixed

BackEnd/chat_notifications/consumer.py
vladmorozov2: fixed

BackEnd/chat_notifications/consumer.py
vladmorozov2: fixed

BackEnd/chat_notifications/consumer.py
vladmorozov2: @mehalyna  If i got approve from you, i will delete this PR and create new with this changes to merge to release branch

mehalyna (approved):

Issue #79, opened 2025-02-26
[Notifications] Send Mails


isac-medison:
Create MVP for sending mails via SMTP based on Kafka Topics

PR #138, opened 2025-03-10
Notification service

isac-medison:
[add] Notifications service consumes and can produce the topic "auth". As a consumer application, it sends emails via SMTP.

* [Main ticket](https://github.com/ITA-Dnipro/Forum_UPD/issues/79)

ToDo
Parallel


BackEnd/notifications-service/.env
NikitaPash, line 1: Your .env file probably wasn't captured by .gitignore

BackEnd/notifications-service/build.sbt
```diff
@@ -0,0 +1,33 @@
...
+      "com.typesafe.akka"       %% "akka-actor-typed"         % akkaVersion,
+      "com.typesafe.akka"       %% "akka-stream"              % akkaVersion,
+      "com.typesafe.akka"       %% "akka-pki"                 % akkaVersion,
+      "org.slf4j"               % "slf4j-nop"                 % "2.0.9",
+      "com.sun.mail"            % "javax.mail"                % "1.6.2",
+      "com.typesafe"            % "config"                    % "1.4.2",
+      "org.scalatra.scalate"    %% "scalate-core"             % "1.10.1",
+      "io.spray"                %% "spray-json"               % "1.3.6",
+      "dev.zio"                 %% "zio"                      % "2.1.15",
+      "dev.zio"                 %% "zio-streams"              % "2.1.15",
```
mehalyna: You are using both Akka Streams (`"com.typesafe.akka" %% "akka-stream"`) and ZIO Streams (`"dev.zio" %% "zio-streams"`). These two libraries serve similar purposes, which may lead to conflicts.

BackEnd/notifications-service/src/main/scala/com/softserve/Main.scala
```diff
@@ -0,0 +1,37 @@
...
+    for {
+      // in need to produce
+      // _ <- AuthMessageProducer.produce(bootstrapServers, topic, message)
+      _ <- AuthMessageConsumer.run(bootstrapServers, templateEngine)
+
+    } yield ()
+  }
+  catch {
+      case ex: Exception => Console.printLine(s"${ex.getMessage}")
+    }
```
mehalyna: You are using `try-catch` in a ZIO application, which does not work properly because ZIO is purely functional and should handle errors with ZIO effects. `catch` block will never execute because `for-comprehension` in ZIO does not throw exceptions in the traditional sense.
Use `ZIO.catchAll` or `ZIO.tapError` for proper error handling.
```
override def run: ZIO[Any, Throwable, Unit] =
  AuthMessageConsumer.run(bootstrapServers, templateEngine)
    .tapError(err => Console.printLine(s"Error: ${err.getMessage}"))
```

BackEnd/notifications-service/src/main/scala/com/softserve/Main.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  override def run: ZIO[Any, Throwable, Unit] = {
+    try {
+     
+    for {
+      // in need to produce
+      // _ <- AuthMessageProducer.produce(bootstrapServers, topic, message)
+      _ <- AuthMessageConsumer.run(bootstrapServers, templateEngine)
+
+    } yield ()
```
mehalyna: Your `for-comprehension` does not return anything (`Unit`). If `AuthMessageProducer.produce` is needed, it should be included properly.
```
override def run: ZIO[Any, Throwable, Unit] =
  (for {
    _ <- AuthMessageConsumer.run(bootstrapServers, templateEngine)
  } yield ())
    .tapError(err => Console.printLine(s"Error: ${err.getMessage}"))
```

BackEnd/notifications-service/src/main/scala/com/softserve/Main.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.fusesource.scalate.TemplateEngine
+import com.softserve.helpers._  
+import com.softserve.constants._  
+import com.softserve.consumers._
+import com.softserve.producers._
+import com.softserve.models._
+import com.softserve.serde._
+
+object Main extends ZIOAppDefault {
+  val bootstrapServers = List("kafka:9092")
```
mehalyna: Hardcoded `kafka:9092` can break deployments where Kafka brokers have different hostnames. Read from environment variables using `System.getProperty` or ZIO Config.
```
val bootstrapServers = sys.env.getOrElse("KAFKA_BROKER", "kafka:9092").split(",").toList
```

BackEnd/notifications-service/src/main/scala/com/softserve/Main.scala
```diff
@@ -0,0 +1,37 @@
...
+  val templateEngine = new TemplateEngine()
+  val topic = TopicConstants.Authentication
+
+  override def run: ZIO[Any, Throwable, Unit] = {
+    try {
+     
+    for {
+      // in need to produce
+      // _ <- AuthMessageProducer.produce(bootstrapServers, topic, message)
```
mehalyna: If needed, uncomment and include this line.  If not needed, remove it to keep the code clean.

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
```diff
@@ -0,0 +1,27 @@
...
+  final val ConfirmEmailSubject: String = "Confirm your e-mail"
+  final val MissingRequiredEnvVariable: String = "Missing required environment variable:"
+  final val requiredEnvVars = Map(
+    "EMAIL_HOST_USER" -> "Email host user",
+    "EMAIL_HOST" -> "Email host",
+    "EMAIL_PORT" -> "Email port",
+    "EMAIL_HOST_PASSWORD" -> "Email host password"
+    )
+    final val envVars: Map[String, String] = requiredEnvVars.map { case (key, description) =>
+      key -> sys.env.getOrElse(key, throw new RuntimeException(MissingRequiredEnvVariable + s"$key ($description)"))
```
mehalyna: Instead of throwing a `RuntimeException`, use ZIO Config or Try to provide graceful error handling.

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
```diff
@@ -0,0 +1,27 @@
...
+    final val ErrorSendingEmail: String = "Error sending email"
+    final val SMTPAuthFailed: String = "SMTP authentication failed"
+    
+  }   
+  object EmailTemplates {
+    final val TemplatesPath: String = "src/main/scala/com/softserve/email_templates"
+    final val ConfirmName: String = "ConfirmEmailTemplate.mustache"
+    final val EventName: String = "EventActionTemplate.mustache"
+    final val ResetPasswordName: String = "ResetPasswordTemplate.mustache"
+  }
```
mehalyna: The `EmailTemplates` object is inside `EmailConstants.scala` but declared separately. This can cause compilation issues.
Move `EmailTemplates` inside `EmailConstants` properly.

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
```diff
@@ -0,0 +1,27 @@
...
+  final val requiredEnvVars = Map(
+    "EMAIL_HOST_USER" -> "Email host user",
+    "EMAIL_HOST" -> "Email host",
+    "EMAIL_PORT" -> "Email port",
+    "EMAIL_HOST_PASSWORD" -> "Email host password"
+    )
+    final val envVars: Map[String, String] = requiredEnvVars.map { case (key, description) =>
+      key -> sys.env.getOrElse(key, throw new RuntimeException(MissingRequiredEnvVariable + s"$key ($description)"))
+    }
+    final val EmailSentSuccessfulyTo: String = "Email sent successfully to"
```
mehalyna: "EmailSentSuccessfulyTo" should be "EmailSentSuccessfullyTo".

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
```diff
@@ -0,0 +1,27 @@
...
+    final val ErrorSendingEmail: String = "Error sending email"
+    final val SMTPAuthFailed: String = "SMTP authentication failed"
+    
+  }   
+  object EmailTemplates {
+    final val TemplatesPath: String = "src/main/scala/com/softserve/email_templates"
```
mehalyna: This may not work in production when deployed inside a container. Use `sys.env.getOrElse("EMAIL_TEMPLATES_PATH", "default_path")` instead.

BackEnd/notifications-service/src/main/scala/com/softserve/constants/FieldConstants.scala
mehalyna, line 1: Scala best practices recommend using `lazy val` for constants to avoid unnecessary initialization at object creation.

mehalyna, line 1: Scala often follows PascalCase for object names but UPPER_CASE for constants. This makes constants easier to identify.

BackEnd/notifications-service/src/main/scala/com/softserve/constants/SystemConstants.scala
mehalyna, line 1: Follow comments to BackEnd/notifications-service/src/main/scala/com/softserve/constants/FieldConstants.scala

BackEnd/notifications-service/src/main/scala/com/softserve/consumers/AuthMessageConsumer.scala
```diff
@@ -0,0 +1,50 @@
...
+import com.softserve.helpers._
+import java.time.LocalDateTime
+
+object AuthMessageConsumer {
+  def run(bootstrapServers: List[String], templateEngine: TemplateEngine): ZIO[Any, Throwable, Unit] = {
+    ZIO.scoped {
+      for {
+        consumer <- Consumer.make(
+          ConsumerSettings(bootstrapServers).withGroupId("notifications-app")
+        )
```
mehalyna, line 20: In ZIO Kafka, `Consumer.make(...)` is deprecated in favor of ZLayer. Using ZLayer allows for better dependency injection and lifecycle management.
Replace `Consumer.make(...)` with:
```
val consumer = ZLayer.scoped(Consumer.make(settings))
```

BackEnd/notifications-service/src/main/scala/com/softserve/consumers/AuthMessageConsumer.scala
```diff
@@ -0,0 +1,50 @@
...
+            val email = authMessage.email
+            val link = authMessage.activationLink
+            val name = authMessage.name
+            val sourceDataPath = new File(s"${EmailTemplates.TemplatesPath}/${EmailTemplates.ConfirmName}").getCanonicalPath
+            val someAttributes = Map("name" -> name, "link" -> link)
+            val temp = templateEngine.layout(sourceDataPath, someAttributes).toString()
+            SmtpEmailSender.sendEmail(
+              List(email),
+              EmailConstants.ConfirmEmailSubject,
+              temp)
```
mehalyna: If an email fails to send, the consumer will continue running without logging or handling errors. This can cause silent failures.
Wrap `SmtpEmailSender.sendEmail(...)` in `ZIO.attempt(...)` and log errors.

BackEnd/notifications-service/src/main/scala/com/softserve/consumers/AuthMessageConsumer.scala
```diff
@@ -0,0 +1,50 @@
...
+              val endTime = LocalDateTime.now()
+            Console.printLine(s"Mail send!")
```
mehalyna: Using `Console.printLine(...)` is not ideal for logging in production systems. Use ZIO Logging for structured logging.

BackEnd/notifications-service/src/main/scala/com/softserve/models/AuthMessage.scala
```diff
@@ -0,0 +1,18 @@
...
+import com.softserve.validations.ValidationUtils
+import zio.json._
+
+case class AuthMessage(email: String, actionType: String, name: String, activationLink: String) {
+  def validate: List[String] = {
+    ValidationUtils.validateEmail(email) ++
+    ValidationUtils.validateNonEmpty(actionType, "ActionType") ++
+    ValidationUtils.validateNonEmpty(name, "Name") ++
+    ValidationUtils.validateNonEmpty(activationLink, "ActivationLink")
+  }
```
mehalyna: If multiple fields are invalid, the validate method returns a list of errors, but there is no clear distinction about which error corresponds to which field. Use case classes for errors instead of plain `List[String]`. Modify `ValidationUtils` to return `List[ValidationError]`, where:
```
case class ValidationError(field: String, message: String)
```

mehalyna: Some fields like `name` and `activationLink` should have length constraints. 
_Example issue:_ A 1-character name might not be valid.
Add a length validation function.

BackEnd/notifications-service/src/main/scala/com/softserve/models/AuthMessage.scala
mehalyna, line 1: There is no logging when validation fails, making debugging harder. Use ZIO Logging to log validation errors.

BackEnd/notifications-service/src/main/scala/com/softserve/producers/AuthMessageProducer.scala
mehalyna, line 1: If Kafka is down or the topic is misconfigured, the application will crash. The function does not catch errors, meaning any failure stops execution. Use ZIO error handling (ZIO.attempt, ZIO.logError, ZIO.retry). Retry failed Kafka message deliveries using an exponential backoff strategy.

mehalyna, line 1: If Kafka is slow, the producer might hang indefinitely. There is no timeout for message sending, meaning a bad network could cause a deadlock. Set a request timeout ("request.timeout.ms") to prevent hanging.
```
val producerSettings = ProducerSettings(bootstrapServers)
  .withProperty("acks", "all")
  .withProperty("retries", "5")
  .withProperty("enable.idempotence", "true")
  .withProperty("request.timeout.ms", "5000") // 5-second timeout for Kafka requests
```

mehalyna (changes requested):

BackEnd/notifications-service/src/main/scala/com/softserve/producers/AuthMessageProducer.scala
isac-medison: The producer is used for testing purposes only, as microservices do not produce notifications yet.  In future we will not have it at all under notifications service.

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
```diff
@@ -0,0 +1,32 @@
...
+  )
+
+object EmailConstants {
+  final lazy val CONFIRM_EMAIL_SUBJECT: String = "Confirm your e-mail"
+  final lazy val MISSING_REQUIRED_ENV_VAR: String = "Missing required environment variable:"
+  final lazy val EMAIL_SENT_SUCCESSFULLY_TO: String = "Email sent successfully to"
+  final lazy val ERROR_SENDING_EMAIL: String = "Error sending email"
+  final lazy val SMTP_AUTH_FAILED: String = "SMTP authentication failed"
+  val emailConfig: EmailConfig = EmailConfig(
+    EMAIL_HOST_USER = sys.env.getOrElse("EMAIL_HOST_USER", throw new RuntimeException("Missing EMAIL_HOST_USER")),
```
IraMladanovych02: The current implementation uses sys.env.getOrElse(...) to retrieve environment variables, which can cause the application to crash if a required variable is missing. Instead of throwing an exception immediately, consider handling missing values gracefully using ZIO.fromOption(...).orElseFail(...)

val emailConfig: Task[EmailConfig] = for {
  user <- ZIO.fromOption(sys.env.get("EMAIL_HOST_USER"))
    .orElseFail(new RuntimeException("Missing EMAIL_HOST_USER"))
  host <- ZIO.fromOption(sys.env.get("EMAIL_HOST"))
    .orElseFail(new RuntimeException("Missing EMAIL_HOST"))
  port <- ZIO.fromOption(sys.env.get("EMAIL_PORT"))
    .orElseFail(new RuntimeException("Missing EMAIL_PORT"))
  password <- ZIO.fromOption(sys.env.get("EMAIL_HOST_PASSWORD"))
    .orElseFail(new RuntimeException("Missing EMAIL_HOST_PASSWORD"))
} yield EmailConfig(user, host, port, password)

val emailConfigLayer: ZLayer[Any, Throwable, EmailConfig] = ZLayer.fromZIO(emailConfig)

IraMladanovych02 (changes requested):

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
isac-medison: To be honest, I don`t see how the service can work without the required variables, which are required for core functionality.
The use of Zio config requires refactoring all methods, which use that config. I propose to discuss that on a daily, And In case needed, to separate that fix into another task

NikitaPash (approved):

BackEnd/notifications-service/build.sbt
isac-medison: Decided to use dev.zio, akka was used for endpoints implementation

BackEnd/notifications-service/src/main/scala/com/softserve/Main.scala
isac-medison: Thanks, resolved, please check

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
mehalyna: The `requiredEnvVars` map does not handle missing environment variables properly. It throws a `RuntimeException` immediately when missing, which could crash the application at startup.
You may use as follows
```
import scala.util.Try

final val envVars: Map[String, String] = requiredEnvVars.map { case (key, description) =>
  key -> Try(sys.env(key)).getOrElse {
    Console.err.println(s"Warning: Missing environment variable: $key ($description)")
    ""
  }
}
```
This prints a warning instead of crashing the app. If an environment variable is missing, it assigns an empty string instead.

mehalyna (changes requested):

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
isac-medison: Is there a need to run the service with empty strings as core parameters? No logic will be executed as intended.

isac-medison: I like the approach Ira suggested in the comment upper.
```
  final lazy val emailConfig: Task[EmailConfig] = for {
    user <- ZIO.fromOption(sys.env.get("EMAIL_HOST_USER")).orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_HOST_USER"))
    host <- ZIO.fromOption(sys.env.get("EMAIL_HOST"))orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_HOST"))
    port <- ZIO.fromOption(sys.env.get("EMAIL_PORT"))orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_PORT"))
    password <- ZIO.fromOption(sys.env.get("EMAIL_HOST_PASSWORD")).orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_HOST_PASSWORD"))
    } yield EmailConfig(user, host, port, password)
```
I still throw exceptions as there is no need to continue service work if we don't have the required vars. But I also add graceful error message

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
isac-medison: Rewrote config, using zio config
```
  final lazy val emailConfig: Task[EmailConfig] = for {
    user <- ZIO.fromOption(sys.env.get("EMAIL_HOST_USER")).orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_HOST_USER"))
    host <- ZIO.fromOption(sys.env.get("EMAIL_HOST"))orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_HOST"))
    port <- ZIO.fromOption(sys.env.get("EMAIL_PORT"))orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_PORT"))
    password <- ZIO.fromOption(sys.env.get("EMAIL_HOST_PASSWORD")).orElseFail(new RuntimeException(s"$MISSING_REQUIRED_ENV_VAR EMAIL_HOST_PASSWORD"))
    } yield EmailConfig(user, host, port, password)
```

mehalyna (approved):

IraMladanovych02 (approved):

PR #140, opened 2025-03-11
116 event microservice management

AntonKovalenk0:
Setup an event microservice with docker compose.
All my changes are in the Backend/event-service folder.
Created:

- main.py for project entry point
- database.py for connecting to PostgreSQL
- models.py and schemas.py for an Event table
- routes/events.py for CRUD endpoints

Updated due to mentor comments:

1. requirements for project.
2. .gitignore
3. change delete to soft_delete

### DIFF
--- a/BackEnd/event-service/app/crud.py
+++ b/BackEnd/event-service/app/crud.py
@@ -1,0 +1,41 @@
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+from fastapi import Depends, HTTPException
+from typing import List, Optional, Type, TypeVar
+from database import StatusEnum
+
+T = TypeVar("T")
+
+async def get_active_objects(db: AsyncSession, model: Type[T], obj_id: int = None, skip: int = 0, limit: int = 10) -> List[T]:
+    active_status_select = select(model).where(model.status != StatusEnum.deleted)
+    if obj_id is not None:
+        active_status_select = active_status_select.where(model.id == obj_id)
+    else:
+        active_status_select = active_status_select.offset(skip).limit(limit)
+    
+    result = await db.execute(active_status_select)
+    if obj_id is not None:
+        return result.scalar_one_or_none()
+    else:
+        return result.scalars().all()
+
+async def soft_delete(db: AsyncSession, model: Type[T], obj_id: int) -> bool:
+    try:
+        obj_to_delete = select(model).where(model.id == obj_id, model.status != StatusEnum.deleted)
+        result = await db.execute(obj_to_delete)
+        obj = result.scalar_one_or_none()
+        
+        if not obj:
+            return False
+        
+        setattr(obj, "status", StatusEnum.deleted)
+
+        await db.commit()
+        return True
+    
+    except Exception:
+        await db.rollback()
+        return False
+
+
+
--- a/BackEnd/event-service/app/crud.py
+++ b/BackEnd/event-service/app/crud.py
@@ -20,16 +20,12 @@ async def get_active_objects(db: AsyncSession, model: Type[T], obj_id: int = Non
         return result.scalars().all()
 
 async def soft_delete(db: AsyncSession, model: Type[T], obj_id: int) -> bool:
+    obj_to_delete = await db.get(model, obj_id)
+    if not obj_to_delete or obj_to_delete.status == StatusEnum.deleted:
+        return False
+    
+    obj_to_delete.status = StatusEnum.deleted
     try:
-        obj_to_delete = select(model).where(model.id == obj_id, model.status != StatusEnum.deleted)
-        result = await db.execute(obj_to_delete)
-        obj = result.scalar_one_or_none()
-        
-        if not obj:
-            return False
-        
-        setattr(obj, "status", StatusEnum.deleted)
-
         await db.commit()
         return True
     
--- a/BackEnd/event-service/app/database.py
+++ b/BackEnd/event-service/app/database.py
@@ -1,0 +1,25 @@
+import os
+from dotenv import load_dotenv
+from enum import Enum
+from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine, async_sessionmaker
+from sqlalchemy.orm import sessionmaker, declarative_base
+
+
+load_dotenv()
+DATABASE_URL = os.getenv("DATABASE_URL_DOCKER")
+print(f"Connecting to database: {DATABASE_URL}")
+
+engine = create_async_engine(DATABASE_URL, echo=True, future=True)
+
+AsyncSessionLocal = async_sessionmaker(bind=engine, expire_on_commit=False)
+
+Base = declarative_base()
+
+class StatusEnum(str, Enum):
+    active = "active"
+    inactive = "inactive"
+    deleted = "deleted"
+
+async def get_async_session() -> AsyncSession:
+    async with AsyncSessionLocal() as session:
+        yield session
--- a/BackEnd/event-service/app/routes/events.py
+++ b/BackEnd/event-service/app/routes/events.py
@@ -1,0 +1,49 @@
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy.ext.asyncio import AsyncSession
+from database import get_async_session, StatusEnum
+from models import Event
+from schemas import EventCreate, EventUpdate
+from crud import get_active_objects, soft_delete
+
+router = APIRouter(prefix="/events", tags=["Events"])
+
+@router.get("/", summary="Get all events")
+async def get_events(skip: int = 0, limit: int = 10, db: AsyncSession = Depends(get_async_session)):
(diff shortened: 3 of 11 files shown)


BackEnd/event-service/app/crud.py
```diff
@@ -0,0 +1,41 @@
...
+        active_status_select = active_status_select.offset(skip).limit(limit)
+    
+    result = await db.execute(active_status_select)
+    if obj_id is not None:
+        return result.scalar_one_or_none()
+    else:
+        return result.scalars().all()
+
+async def soft_delete(db: AsyncSession, model: Type[T], obj_id: int) -> bool:
+    try:
```
VovaD199, line 28: select(model).where(...).execute() executes the query twice: once for the selection, the second time for the update.

async def soft_delete(db: AsyncSession, model: Type[T], obj_id: int) -> bool:
    obj = await db.get(model, obj_id)
    
    if not obj or obj.status == StatusEnum.deleted:
        return False

    obj.status = StatusEnum.deleted  # Оновлюємо статус

    try:
        await db.commit()
        return True
    except Exception:
        await db.rollback()
        return False

BackEnd/event-service/app/routes/events.py
```diff
@@ -0,0 +1,49 @@
...
+    return {"message": "Event created successfully", "event": event}
+
+@router.get("/{event_id}", summary="Get an event")
+async def get_event(event_id: int, db: AsyncSession = Depends(get_async_session)):
+    event = await get_active_objects(db=db, model=Event, obj_id=event_id)
+    if event is None:
+        raise HTTPException(status_code=404, detail="Event not found")
+    return event
+
+@router.delete("/events/{event_id}", summary="Delete an event")
```
VovaD199: @router.delete("/events/{event_id}") → @router.delete("/{event_id}")

BackEnd/event-service/app/routes/events.py
```diff
@@ -0,0 +1,49 @@
...
+async def delete_event(event_id: int, db: AsyncSession = Depends(get_async_session)):
+    success = await soft_delete(db=db, model=Event, obj_id=event_id)
+    if not success:
+        raise HTTPException(status_code=404, detail="Event not found")
+    return {"message": "Event deleted successfully"}
+
+@router.put("/events/{event_id}", summary="Update an event")
```
VovaD199: @router.put("/events/{event_id}") → @router.put("/{event_id}")

BackEnd/event-service/app/routes/events.py
```diff
@@ -0,0 +1,49 @@
...
+
+@router.get("/", summary="Get all events")
+async def get_events(skip: int = 0, limit: int = 10, db: AsyncSession = Depends(get_async_session)):
+    events = await get_active_objects(db=db, model=Event, skip=skip, limit=limit)
+    return events
+
+@router.post("/", summary="Create an event")
+async def create_event(event: EventCreate, db: AsyncSession = Depends(get_async_session)):
+    db_event = Event(**event.dict())
+    db_event.organizer_id = 1
```
VovaD199, line 18: db_event.organizer_id = 1 (must be taken from the event or authorized user).

BackEnd/event-service/app/routes/events.py
```diff
@@ -0,0 +1,49 @@
...
+    db.add(db_event)
+    await db.commit()
+    await db.refresh(db_event)
+    return {"message": "Event created successfully", "event": event}
```
VovaD199: Incorrect return in create_event()
We need to return the saved db_event object.

BackEnd/event-service/app/schemas.py
```diff
@@ -0,0 +1,27 @@
+from pydantic import BaseModel
+from typing import Optional
+from datetime import date, time
+
+class EventCreate(BaseModel):
+    title: str
+    description: Optional[str] = None
+    date: date
+    time: time
```
VovaD199: The date and time must either be removed or made so that they cannot be empty, because it will look ugly an event without a date and time

BackEnd/event-service/app/schemas.py
```diff
@@ -0,0 +1,27 @@
...
+    location: Optional[str] = None
+    capacity: int
+    available_slots: int
```
VovaD199, line 12: Can these parameters be None?

VovaD199 (changes requested):

BackEnd/event-service/app/routes/events.py
AntonKovalenk0: I decided to set this parameter as 1 because I can't receive user ID from the token.  I'll change it when we can communicate with authentication service.

BackEnd/event-service/app/routes/events.py
AntonKovalenk0: Agreed. Thanks.

BackEnd/event-service/app/crud.py
AntonKovalenk0: I agree, that your method is more efficient and readable. I'll read more about db.get()

BackEnd/event-service/app/routes/events.py
AntonKovalenk0: Yes, it`s my inattention

BackEnd/event-service/app/schemas.py
AntonKovalenk0: The date and time fields are required in criating an event and I changed EventUpdate class also

BackEnd/event-service/app/schemas.py
AntonKovalenk0: I changed EventUpdate schema so all fields are required and notOptional

AntonKovalenk0: About None, I decided to set: description, location, image, summary fields as Optional so they're None by default. We can update description, location. And I'll add images and summary routes later.

VovaD199 (approved):

BackEnd/event-service/app/routes/events.py
vladmorozov2, line 18: In my opinion, it might be better to receive the organizer ID from the POST method rather than from JWT, as I’m not sure if you should be the one handling this. I believe this responsibility falls under the Sofia and Iryna services. @VovaD199 , what do you think?

AntonKovalenk0: I think when we are creating an event, we might set an organizer_id. It will be the same id that is contained in other service's table. In my table I store only INT value. About JWT, it will contain info about authorized user in current session so I can communicate with other service and receive this ID.  Right now as we aren't communicating with other services, I set default id so I can create an event.

AntonKovalenk0: Can you explain what exectly you propose to change now, because I don't quite understand. @vladmorozov2

vladmorozov2: In my opinion it not your responsibility to think about jtw. You only receive validated request and it handle it not jwt
https://chatgpt.com/share/67d1f025-b1c4-800b-91ca-2a286c83435a

AntonKovalenk0: Ok agreed. So how can I receive an ID from other service's table?

vladmorozov2: I think you could receive parameters (organizer id ) in post request

vladmorozov2: I think it's more of a proposal to discuss rather than a remark. I'll bring it up in the daily meeting.

AntonKovalenk0: I think it is also valid idea, but we can ask about this moment an expert on meeting, or discuss it together.

vladmorozov2 (approved):
Good job!

BackEnd/event-service/app/routes/events.py
hrybvalerii: But how can user know their id(without info encoded in jwt) to put it into post request?

vladmorozov2: in ideal world I think it should send frontend which know what user id

BackEnd/event-service/app/database.py
```diff
@@ -0,0 +1,25 @@
+import os
+from dotenv import load_dotenv
+from enum import Enum
+from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine, async_sessionmaker
+from sqlalchemy.orm import sessionmaker, declarative_base
+
+
+load_dotenv()
+DATABASE_URL = os.getenv("DATABASE_URL_DOCKER")
+print(f"Connecting to database: {DATABASE_URL}")
```
hrybvalerii, line 10: I think it`s better to remove this print

hrybvalerii (approved):

BackEnd/event-service/app/database.py
AntonKovalenk0: Agree, it's for development

PR #153, opened 2025-03-16
Task/152 docker change postgres services

AntonKovalenk0:
Setup postgres image for to connect for all needed microservices.
Create docker-compose.services.yml file to add all services to run together.
- inside init.sh file, that contains script for creating multiple postgres databases, changed creating different users for databases to creating one user and granting accesses
To run this file inside docker we need to change Newline style from CRLF(Windows) to LF(Linux) for this file. Thats will able to create multiple databases.
- inside docker-compose.db.yml file, made little changes.
- created docker-compose.services.yml file where all microservices will run together. Added personal event service there.

### DIFF
--- a/docker-compose.services.yml
+++ b/docker-compose.services.yml
@@ -4,7 +4,22 @@ services:
     event-api:
       container_name: event_api
       build: ./Backend/event-service
-      command: bash -c 'uvicorn app.main:app --host 0.0.0.0 --reload'
+      command: >
+        bash -c "
+        while ! </dev/tcp/db/5432; do 
+          echo 'Waiting for PostgreSQL to be available...'; 
+          sleep 1; 
+        done; 
+        echo 'PostgreSQL is ready!';
+        while ! </dev/tcp/redis/6379; do 
+          echo 'Waiting for Redis to be available...'; 
+          sleep 1; 
+        done; 
+        echo 'Redis is ready!';
+        gunicorn app.main:app 
+        --workers 4 
+        --worker-class uvicorn.workers.UvicornWorker 
+        --bind 0.0.0.0:8000"
       volumes:
         - ./BackEnd/event-service:/src
       ports:
--- a/docker-compose.services.yml
+++ b/docker-compose.services.yml
@@ -1,0 +1,24 @@
+version: '3.8'
+
+services:
+    event-api:
+      container_name: event_api
+      build: ./Backend/event-service
+      command: bash -c 'uvicorn app.main:app --host 0.0.0.0 --reload'
+      volumes:
+        - ./BackEnd/event-service:/src
+      ports:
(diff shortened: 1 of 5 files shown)


docker-compose.services.yml
```diff
@@ -0,0 +1,24 @@
+version: '3.8'
+
+services:
+    event-api:
+      container_name: event_api
+      build: ./Backend/event-service
+      command: bash -c 'uvicorn app.main:app --host 0.0.0.0 --reload'
+      volumes:
+        - ./BackEnd/event-service:/src
```
mehalyna: The service mounts `./BackEnd/event-service:/src`, but inside the container, you try to run `uvicorn` from `/src/app`. This does not match the volume path, which might cause `ModuleNotFoundError`.

docker-compose.services.yml
```diff
@@ -0,0 +1,24 @@
+version: '3.8'
+
+services:
+    event-api:
+      container_name: event_api
+      build: ./Backend/event-service
+      command: bash -c 'uvicorn app.main:app --host 0.0.0.0 --reload'
```
mehalyna: The `--reload` flag is only useful in development. In production, Uvicorn should be run without `--reload` and use multiple workers. Replace `--reload` with proper Gunicorn-based Uvicorn workers.
```
command: >
  gunicorn app.main:app 
  --workers 4 
  --worker-class uvicorn.workers.UvicornWorker 
  --bind 0.0.0.0:8000
```

docker-compose.services.yml
mehalyna, line 1: No health check for the database or cache. Services might start too soon before dependencies are ready.

mehalyna (changes requested):

docker-compose.services.yml
AntonKovalenk0: I have WORKDIR /src inside Dockerfile, so all my code is inside /src folder while container is running. 
image
then I run my "main.py" file, which contains inside "app/" folder. It doesn't cause ModuleNotFoundError when starting my service.
I also set PYTHONPATH=/src/app  so I can import from my "app" folder where I'm working

docker-compose.services.yml
AntonKovalenk0: Agree with your comment.

docker-compose.services.yml
AntonKovalenk0: Inside docker-compose.db.yml I added healthcheck for redis and postgreSQL. Also for my event-api container added special commands to wait till we can connect to </dev/tcp/db/5432 and </dev/tcp/redis/6379. And then start FastAPI, ensuring dependencies are ready

sofiahnatiuk (approved):

mehalyna (approved):

hrybvalerii (approved):

PR #156, opened 2025-03-17
[News] 137 add redis cache and implement periodic task

Maksym04u:
dev
## JIRA

## Code reviewers

- [ ] @github_username

### Second Level Review

- [ ] @github_username

## Summary of issue

ToDo

## Summary of change

ToDo

## Testing approach

ToDo

## CHECK LIST
- [ ]  СI passed
- [ ]  Сode coverage >=95%


BackEnd/news-service/app/celery.py
```diff
@@ -1,17 +1,36 @@
...
+)
+
+# Define a periodic task
+entry = RedBeatSchedulerEntry(
+    "scrape_news_task",
+    "app.tasks.scrape_news_task",
+    crontab(minute=0, hour='*/3'),  # Runs every 3 hours
+    app=celery,
+)
+entry.save()
```
mehalyna: The `RedBeatSchedulerEntry.save()` is called at module level. This means that every time the module is imported (e.g., when Celery starts), a new `RedBeat` entry might be created or overridden, which is not ideal.
Move `entry.save()` inside a function that is only executed when necessary (e.g., as part of an initialization script).

BackEnd/news-service/app/celery.py
```diff
@@ -1,17 +1,36 @@
...
+    backend=REDIS_URL_BACKEND,
     include=["app.tasks"]
 )
 
+celery.conf.update(
+    {
+        "beat_scheduler": "redbeat.RedBeatScheduler",  # Use RedBeat scheduler
+        "redbeat_redis_url": REDIS_URL_BROKER,  
+        "timezone": "UTC",
+    }
```
mehalyna: You should explicitly define task serialization and result serialization to avoid compatibility issues.
```
celery.conf.update(
    {
        "beat_scheduler": "redbeat.RedBeatScheduler",
        "redbeat_redis_url": REDIS_URL_BROKER,
        "timezone": "UTC",
        "task_serializer": "json",  # Recommended for compatibility
        "result_serializer": "json",
        "accept_content": ["json"],  # Ensure Celery accepts JSON payloads only
        "worker_concurrency": 2,  # Set concurrency level (adjust based on workload)
    }
)
```

BackEnd/news-service/app/celery.py
```diff
@@ -1,17 +1,36 @@
...
+from redbeat import RedBeatSchedulerEntry
 from dotenv import load_dotenv
 
 load_dotenv()
 
 
-# Load environment variables
-REDIS_URL = os.getenv("REDIS_URL", "redis://redis_cache:6379/0")
+REDIS_URL_BROKER = os.getenv("REDIS_URL_BROKER", "redis://redis_cache:6379/0")
+REDIS_URL_BACKEND = os.getenv("REDIS_URL_BACKEND", "redis://redis_cache:6379/1")
```
mehalyna: Backend and broker are using different Redis databases (/0 and /1). Are you sure?

BackEnd/news-service/app/routes.py
```diff
@@ -20,50 +23,37 @@ async def get_news(skip: int = 0, limit: int = 3):
...
-        existing = await NewsModel.find_one({"link": news.link})
-        if existing:
-            skipped_news.append(news)
-            continue  
+@router.get("/news/recent/")
+async def get_recent_news():
+    cached_news = await redis.get("recent_news")
+    if cached_news:
+        return json.loads(cached_news)
+    return {"message": "No recent news found in cache."}
```
mehalyna: The function directly calls `redis.get("recent_news")`, but Redis might be down or return `None`. Add error handling if possible.

BackEnd/news-service/app/routes.py
```diff
@@ -20,50 +23,37 @@ async def get_news(skip: int = 0, limit: int = 3):
...
-        "message": "Scraping completed",
-        "saved_news_count": len(saved_news),
-        "skipped_news_count": len(skipped_news),
-        "saved_news": saved_news,
-        "skipped_news": skipped_news
-    }
+@router.post("/scrape/")
+async def trigger_scraping():
+    """Trigger the Celery task to scrape news"""
+    task = celery.send_task("app.tasks.scrape_news_task")  # Reference the task in tasks.py
```
mehalyna: You use a string reference. This is error-prone if the task function is renamed. Import the actual Celery task and call it directly.

BackEnd/news-service/app/scraper.py
```diff
@@ -112,4 +113,37 @@ async def scrape_article(session, article_url):
...
+        existing = await NewsModel.find_one({"link": news.link})
+        if existing:
+            skipped_news.append(news)
+            continue
+        inserted = await NewsModel.insert_one(news)
+        if inserted.id:
+            saved_news.append(news)
+
+    if saved_news:
+        await update_news_cache(saved_news)
```
mehalyna: The function checks for duplicates one by one (`find_one({"link": news.link})`). This results in N separate database queries (slow if many articles). Fetch all existing links in one query and store them in a set.

BackEnd/news-service/app/scraper.py
```diff
@@ -112,4 +113,37 @@ async def scrape_article(session, article_url):
...
+    skipped_news = []
+
+    for news in latest_news:
+        existing = await NewsModel.find_one({"link": news.link})
+        if existing:
+            skipped_news.append(news)
+            continue
+        inserted = await NewsModel.insert_one(news)
+        if inserted.id:
+            saved_news.append(news)
```
mehalyna: The function inserts news one at a time, causing multiple I/O operations. This is inefficient in databases like MongoDB.
Use bulk inserts for better performance:
```
await NewsModel.insert_many(saved_news)
```

BackEnd/news-service/app/scripts/delete_news.lua
```diff
@@ -0,0 +1,77 @@
...
+redis.log(redis.LOG_NOTICE, "Successfully decoded JSON, total articles: " .. tostring(#news_list))
+
+local updated_news_list = {}
+local found = false
+
+for _, article in ipairs(news_list) do
+    redis.log(redis.LOG_NOTICE, "Full article JSON: " .. cjson.encode(article))
+
+    if type(article) == "table" then
+        local article_id = tostring(article["id"])
```
mehalyna: The script assumes that every article has an `id` field, but this may not always be true. Check if `article["id"]` exists before attempting to convert it.

BackEnd/news-service/app/scripts/update_news.lua
```diff
@@ -0,0 +1,22 @@
...
+end
+
+table.insert(news_list, 1, new_news)
+
+while #news_list > max_size do
+    table.remove(news_list)
+end
+
+
+redis.call("SETEX", key, ttl, cjson.encode(news_list))
```
mehalyna: The TTL is fixed at 57600 seconds (~16 hours). If the key already has a TTL, overwriting it might not be ideal.
Retrieve the existing TTL instead of hardcoding:
```
local ttl = redis.call("TTL", key)
if ttl <= 0 then ttl = 57600 end
```

BackEnd/news-service/app/scripts/update_news.lua
```diff
@@ -0,0 +1,22 @@
...
+local new_news = cjson.decode(ARGV[1])
+local max_size = tonumber(ARGV[2])
+local ttl = 57600
+
+local existing_news = redis.call("GET", key)
+local news_list = {}
+
+if existing_news then
+    news_list = cjson.decode(existing_news)
+end
```
mehalyna: If `existing_news` contains corrupt JSON, `cjson.decode(existing_news)` will fail. The script does not handle decoding errors and could crash Redis. Wrap JSON decoding in `pcall()` to catch errors.

BackEnd/news-service/app/scripts/update_news.lua
mehalyna, line 1: The script decodes the entire JSON list into Lua, inserts an item, and encodes it again. This can be slow for large lists.
Use Redis LIST commands (LPUSH & LTRIM) for efficiency.

BackEnd/news-service/app/tasks.py
```diff
@@ -1 +1,23 @@
...
+        asyncio.set_event_loop(loop)
+
+        loop.run_until_complete(init_db())
+
+        loop.run_until_complete(scrape_and_store_news())
+    except Exception as e:
+        logger.error(f"Error in scrape_news_task: {e}", exc_info=True)
+        raise self.retry(exc=e)  
+    finally:
+        loop.close()
```
mehalyna: The script creates a new event loop and sets it as the current event loop. This approach can cause issues because Celery tasks do not run in an actual event loop. Creating and closing event loops improperly can lead to resource leaks and performance degradation.
Instead of manually creating an event loop, use `asyncio.run()`, which is safer:
```
@celery.task(bind=True)
def scrape_news_task(self):
    """Scrapes and stores news asynchronously"""
    try:
        asyncio.run(init_db())
        asyncio.run(scrape_and_store_news())
    except Exception as e:
        logger.error(f"Error in scrape_news_task: {e}", exc_info=True)
        raise self.retry(exc=e)
```

BackEnd/news-service/app/tasks.py
```diff
@@ -1 +1,23 @@
...
+    try:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+
+        loop.run_until_complete(init_db())
+
+        loop.run_until_complete(scrape_and_store_news())
+    except Exception as e:
+        logger.error(f"Error in scrape_news_task: {e}", exc_info=True)
+        raise self.retry(exc=e)
```
mehalyna: `self.retry(exc=e)` is used without proper Celery retry settings. This means the task will fail instantly instead of retrying automatically.
Add `autoretry_for` and retry delays.

BackEnd/news-service/app/utils.py
```diff
@@ -0,0 +1,42 @@
...
+            return str(obj)
+        if isinstance(obj, datetime):
+            return obj.isoformat()
+        return super().default(obj)
+    
+
+def load_lua_script(path):
+    """Loads the Lua script for Redis caching."""
+    print("enter lua")
+    with open(path, "r") as file:
```
mehalyna: No exception handling is included, so the script will crash if the file is missing.

BackEnd/news-service/app/utils.py
mehalyna, line 1: Print statements (`print("entered")`, `print(result)`) will not work well in production since print statements do not get logged into files.
Use logging instead

mehalyna (changes requested):

NikitaPash (approved):
Fix Halyna's remarks and it will be ready, good job.

BackEnd/news-service/app/celery.py
Maksym04u: I actually thought that it might be needed to use different databases. I thought one will be used as broker the other one as cache. But I now read that there is no need to do that and better to keep it simple, I will use only one db. Thanks for the remark.

BackEnd/news-service/app/scraper.py
Maksym04u: At most there will be like 10 news so 10 queries. Moreover field "**link**" is **unique**, so looking for it won't take much time. And if we'll be fetching and storing in set all our news it might be a problem, cause we can have like millions news in db. That would be overwhelming and not effective.

BackEnd/news-service/app/scripts/update_news.lua
Maksym04u: Actually it is how I want it to work. We have cache which will be updating its ttl everytime we adding some new news to cache. If I implemented it the way you suggest, there would be high possibility of cache key **ttl run out** and our server would be **without cache** at all till the next scrape. It would be a bad thing to happen

BackEnd/news-service/app/tasks.py
Maksym04u: I already tried it. But got problem **RuntimeError: Event loop is closed**. And according to the documentation [https://docs.python.org/3.10/library/asyncio-task.html#asyncio.run](url) asyncio.run : 
_This function cannot be called when another asyncio event loop is running in the same thread._
I suppose that Celery also creates an event loop for yourself, in which we trying to run  **asyncio.run()**. That highly likely is a problem. To solve this I create here another event loop dedicated directly to asyncio. This might not be so effective and simple but seems to be the only way.

BackEnd/news-service/app/scraper.py
mehalyna: A batch query for only the relevant links would be more efficient—fetch existing links in one query, store them in a set, and check against it. This reduces queries from N to 1 while avoiding fetching millions of records.

BackEnd/news-service/app/tasks.py
mehalyna: You're correct that Celery already runs in an event loop, making `asyncio.run()` problematic. Instead of manually managing event loops, try using loop.run_until_complete() on a retrieved loop:
```
@celery.task(bind=True)
def scrape_news_task(self):
    """Scrapes and stores news asynchronously"""
    try:
        loop = asyncio.get_event_loop()
        if loop.is_closed():
            loop = asyncio.new_event_loop()
            asyncio.set_event_loop(loop)

        loop.run_until_complete(init_db())
        loop.run_until_complete(scrape_and_store_news())

    except Exception as e:
        logger.error(f"Error in scrape_news_task: {e}", exc_info=True)
        raise self.retry(exc=e)
```
This avoids conflicts with Celery’s internal loop while keeping tasks async-friendly.

mehalyna (approved):

PR #165, opened 2025-03-20
Profile service#42

hrybvalerii:
dev
## JIRA

## Code reviewers

- [ ] @github_username

### Second Level Review

- [ ] @github_username

## Summary of issue

ToDo

## Summary of change

added model for investor

refactored profile repository


BackEnd/profiles_service/crud/profiles.py
VovaD199, line 1: Uses profile.__dict__.update(profile_dict), which can change service fields (better to use setattr).

No rollback in case of error – if an error occurs after commit(), data can be partially updated without recovery.

Code in ProfileStartupRepository is duplicated – logic for working with categories and regions can be moved to a separate method.

BackEnd/profiles_service/models/profiles.py
VovaD199, line 1: Lack of ForeignKey for relationship() – relationships must be explicitly specified.

Problems with updated_at – it does not update automatically when a record changes.

VovaD199 (changes requested):

BackEnd/profiles_service/crud/profiles.py
hrybvalerii: Is it better to create a separate 2 methods: one for fetching categories and another for regions or one that will handle all many to many fields?

BackEnd/profiles_service/models/profiles.py
hrybvalerii: I will add  onupdate=func.now() then
But Foreign key constraints are specified in auxillary tables that handle many-to-many relasionship, is it necessary to  include them into the original table as well?

BackEnd/profiles_service/crud/profiles.py
hrybvalerii: I changed dict.update to setattr and moved categories and regions fetching to a separate method. 
Could you please explain how data can be only partially updated after commit(), because I dont particularly understand what types of errors need to be handled.

VovaD199: async def add_one(self, profile_dict: dict):
    try:
        profile_dict["is_deleted"] = False
        profile = self.model(**profile_dict)
        self.session.add(profile)
        await self.session.commit()
        return profile
    except Exception as e:
        await self.session.rollback()
        raise e

BackEnd/profiles_service/models/profiles.py
VovaD199: > I will add onupdate=func.now() then But Foreign key constraints are specified in auxillary tables that handle many-to-many relasionship, is it necessary to include them into the original table as well?

In SQLAlchemy, when you define relationships between tables (relationship()), the ORM needs to know which fields are used to relate the tables. If ForeignKey is not explicitly specified, the following problems can occur:

SQLAlchemy does not know how to create a correct JOIN

If ForeignKey is not specified, SQLAlchemy does not know which field to use to join (JOIN) the related table.
This can cause errors in selections or incorrect results.
Automatic table creation (Alembic, create_all()) may not be correct

If ForeignKey is not specified, then the database will not receive foreign key constraints during schema generation, and you will not be able to ensure data integrity.
Deleting or updating records will not cascade

If Foreign Key is not specified, then when deleting the related record, SQLAlchemy will not be able to delete or update the corresponding data automatically.
For example, if you delete a category, SQLAlchemy will not understand whether it needs to be deleted from startup_profiles.

example:
profile_categories: Mapped[list["StartupCategoryOrm"]] = relationship(
    "StartupCategoryOrm",
    secondary="startup_profile_category",
    back_populates="startup_category_profiles",
    cascade="all, delete"
)

BackEnd/profiles_service/crud/profiles.py
VovaD199: > No rollback in case of error – if an error occurs after commit(), data can be partially updated without recovery.

But why isn’t the one automatic call to ROLLBACK enough? Why must I ROLLBACK again?
The rollback that’s caused by the flush() is not the end of the complete transaction block; while it ends the database transaction in play, from the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) point of view there is still a transaction that is now in an inactive state.

[https://docs.sqlalchemy.org/en/20/faq/sessions.html#but-why-isn-t-the-one-automatic-call-to-rollback-enough-why-must-i-rollback-again](url)

BackEnd/profiles_service/models/profiles.py
hrybvalerii: But in case of many-to-many the association table (startup_profile_category) handles the relationship by having 2 foreign keys referencing both original tables. 
For example here foreign keys are used just in an auxillary table: https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#many-to-many

VovaD199: > But in case of many-to-many the association table (startup_profile_category) handles the relationship by having 2 foreign keys referencing both original tables. For example here foreign keys are used just in an auxillary table: https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#many-to-many

I agree, the issue is resolved.

VovaD199 (approved):

PR #167, opened 2025-03-23
Task#99 Customize JWT with user roles

sofiahnatiuk:
1. Added JWT customization
2. Modified login process to let users choose if they want to login as startups or investors
3. Updated tests for the new login process
4. Git marked files as renamed, because I moved them to BackEnd/authentication_service/ directory for consistency with other services


BackEnd/authentication_service/custom_auth/factories.py
```diff
@@ -14,13 +14,12 @@ class Meta:
     password = factory.Faker("password")
 
     is_active = True
```
VovaD199, line 16: why did you remove these fields, a faker can generate them with random values ​​from a list:
    is_startup = False
    is_investor = False
    is_startup_valid = False
    is_investor_valid = False

BackEnd/authentication_service/custom_auth/models.py
```diff
@@ -0,0 +1,60 @@
...
+
+    user = models.ForeignKey("CustomUser", on_delete=models.CASCADE, related_name="user_roles")
+    role = models.ForeignKey("Role", on_delete=models.CASCADE)
+    status = models.CharField(max_length=20, choices=STATUS_CHOICES, default=NOT_VALIDATED)
+
+    class Meta:
+        unique_together = ("user", "role")  # Prevents having the same user with the same role twice
+
+    def __str__(self):
+        return f"{self.user.email} - {self.role.name} ({self.status})"
```
VovaD199, line 60: If the user role is verified, the account can be automatically activated.

    def save(self, *args, **kwargs):
        super().save(*args, **kwargs)
        if self.status == UserRoleStatus.VALIDATED:
            self.user.is_active = True
            self.user.save(update_fields=["is_active"])

VovaD199 (changes requested):

BackEnd/authentication_service/custom_auth/factories.py
sofiahnatiuk: I removed these fields from the factory because I removed them from CustomUser model. Instead, I'm using Role and UserRole models, which is more effective. 
I decided to use these models instead of fields like is_startup, is_startup_valid, is_investor, is_investor_valid in CustomUser model, because the field is_startup_valid depends on is_startup (and the same with investors), which does not satisfy  normal forms of the database

BackEnd/authentication_service/custom_auth/models.py
sofiahnatiuk: An account can be validated only after the user fills out the profile information in the Profile service. To edit a profile, the user needs to log in. is_active means that the user has activated their email account and is allowed to log in (users aren't alowed to log in before email verification). So, before a role is verified (before the user becomes a validated startup or investor), the user must confirm the email, which will set is_active to True

IraMladanovych02 (approved):

VovaD199 (approved):

PR #168, opened 2025-03-24
Feature/146 event registration

AntonKovalenk0:
Completed task 146.
1.Refactored previous structure: moved events managing logic to routes/events/ folder
2. Added Alembic migrations
3. Created routes/participants folder for registration logic and implemented all needed endpoints
4. Minor changes in docker-compose with container names and main file.
5. Separated Pydentic basyc_schema.py to fix circular imports.
6. Rewrited previous SQLAlchemy  models code to 2.0 standarts.


NikitaPash (approved):
Seems good

IraMladanovych02 (approved):

BackEnd/event-service/app/routes/events/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Event(Base):
+    __tablename__ = "events"
+    
+    id: Mapped[intpk]
+    title: Mapped[str_100]
+    description: Mapped[str | None]
+    organizer_id: Mapped[int]
+    date: Mapped[datetime.date] = mapped_column(Date)
+    starting_time: Mapped[datetime.time] = mapped_column(Time)
```
VovaD199, line 17: Add validation starting_time
Now starting_time can be anything.
It is better to check that it is not in the past:

@field_validator("starting_time", mode="before")
def validate_starting_time(cls, v, info):
    event_date = info.data.get("date")
    if event_date and datetime.combine(event_date, v) < datetime.now():
        raise ValueError("Starting time cannot be in the past.")
    return v

VovaD199 (approved):

BackEnd/event-service/app/routes/events/models.py
AntonKovalenk0: Valid point. I will do it with my next PR (categories and enhansing functionality)

PR #171, opened 2025-03-24
Profile service#42

hrybvalerii:
dev
## JIRA

## Code reviewers

- [ ] @github_username

### Second Level Review

- [ ] @github_username

## Summary of issue

ToDo

## Summary of change

* Created services package for business llogic
* Added endpoints for crud operations with investors
* Implemented validation for category and region uniqueness
* Implemented service creation via Depends

## CHECK LIST
- [ ]  СI passed
- [ ]  Сode coverage >=95%


NikitaPash (approved):
Looks nice to me

sofiahnatiuk (approved):

BackEnd/profiles_service/routers/investors.py
```diff
@@ -0,0 +1,100 @@
...
+            status_code=404, detail=f"{e}"
+            )
+    except InvalidRelatedEntityError as e:
+        raise HTTPException(
+            status_code=400, detail=f"{e}"
+            )
+    return profile
+
+
+@router.delete("/{profile_id}")
```
VovaD199, line 89: Why and for what purpose is the Investor Profile deleted? In my opinion, it is better to implement a soft deletion.

BackEnd/profiles_service/utils/repositories.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+    async def get_by_id(self, instance_id: int):
+        query = self._get_query().where(self.model.id == instance_id)
+        for field in self.many_to_many:
+            query = query.options(selectinload(getattr(self.model, field)))
+        result = await self.session.execute(query)
+        profile = result.scalars().first()
+        if not profile:
+            raise NotFoundError("Object not foud")
```
VovaD199: Error in get_by_id() (typo in exception)
"Object not foud" → "Object not found"

BackEnd/profiles_service/utils/repositories.py
```diff
@@ -0,0 +1,123 @@
...
+    def _get_query(self):
+        return select(self.model).where(self.model.is_deleted == False)
+    
+
+    async def add_one(self, profile_dict: dict):
+        profile_dict["is_deleted"] = False
+        return await super().add_one(profile_dict)
+
+
+    async def soft_delete(self, profile_id: int):
```
VovaD199, line 116: If profile_id is not found, commit() is still executed.
Better: Use raise NotFoundError() before try-except.

Example:
async def soft_delete(self, profile_id: int):
    profile = await self.get_by_id(profile_id)
    if not profile:
        raise NotFoundError(f"Profile {profile_id} not found")

    try:
        profile.is_deleted = True
        await self.session.commit()
    except Exception as e:
        await self.session.rollback()
        raise e

VovaD199 (changes requested):

BackEnd/profiles_service/routers/investors.py
hrybvalerii: It actually performs a soft delete under the hood. Should I rename the router function and service method to avoid confusion?

BackEnd/profiles_service/utils/repositories.py
hrybvalerii: My get_by_id method is implemented in such a way that it will raise NotFoundError when object with such id does not exist, so commit() shouldn`t be executed.

    async def get_by_id(self, instance_id: int):
        query = self._get_query().where(self.model.id == instance_id)
        for field in self.many_to_many:
            query = query.options(selectinload(getattr(self.model, field)))
        result = await self.session.execute(query)
        profile = result.scalars().first()
        if not profile:
            raise NotFoundError("Object not found")
        return profile

VovaD199 (approved):

PR #172, opened 2025-03-25
111-implement-search-with-elasticsearch

NikitaPash:
This PR adds all search-related endpoints with nessesary search and filtering logic.

Search is divided into two parts: global search and local (service/index specific) search.

-Global search performs basic searching on main services' topics ;
-Service/index scecific search supports different types of filtering and sorting for the records;

Search query supports matches, that are similar, but not identical to the text in search fields.

All responses support pagination and are cached with redis (cache lives 120 sec).
Suggestions can be enabled to every endpoint.

To demonstrate search capabilities of the service, temprorary ES documents' seeding was provided.


BackEnd/search-service/app/routes/search.py
mehalyna, line 1: Short TTLs are fine for highly dynamic content, but some types (like event filtering or news search) might benefit from longer cache durations if data doesn’t change often. Parameterize TTL based on endpoint or query content.

BackEnd/search-service/app/routes/search.py
```diff
@@ -1,57 +1,326 @@
...
+            result[hit.meta.index].append(result_item)
+
+    except Exception as e:
+        logger.error(f"Error executing Elasticsearch search: {e}", exc_info=True)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Search service is currently unavailable. Please try again later."
+        )
+
+    final_result = {"results": result}
```
mehalyna: Currently, `global search` returns grouped results. For better UX, consider adding a `total_hits` or per-index count field:
```
final_result = {
    "results": result,
    "total_hits": response.hits.total.value
}
```

BackEnd/search-service/app/routes/search.py
```diff
@@ -1,57 +1,326 @@
...
-                            service_name: str,
-                            es_client=Depends(get_es_client)
-                            ):
-    """Service-specific search endpoint."""
-    if service_name not in VALID_SERVICES:
-        logger.error(f"Invalid service name attempted: {service_name}")
-        raise HTTPException(status_code=400, detail=f"Invalid service name: {service_name}")
+async def global_search(
+        request: Request,
+        q: str = Query(..., description="Query string for the search"),
```
mehalyna: Protect against abuse or heavy queries by limiting input length:
```
q: str = Query(..., min_length=2, max_length=100)
```

BackEnd/search-service/app/routes/search.py
mehalyna, line 1: You're using `fields=["title^2", "content"]` in multiple places. Consider defining this in config or processor classes for reuse.

mehalyna, line 1: If you notice the same search terms frequently being used, consider precomputing or storing suggestions separately to avoid computing them on-the-fly.

BackEnd/search-service/app/search_processing/local_search_params.py
mehalyna, line 1: Using `str` for fields like `date_from`, `created_from`, `published_from` works but lacks validation.
```
from datetime import date
date_from: Optional[date] = Field(None, description="Filter events occurring after this date")
```

mehalyna, line 1: Currently, `sort_order` is validated via regex. This is fragile and error-prone.
```
from enum import Enum

class SortOrder(str, Enum):
    asc = "asc"
    desc = "desc"

# Then use:
sort_order: SortOrder = Field(SortOrder.desc, description="Sort order")
```
Same applies to:
- `event_type`
- `event_status`
- `status` (in questions)

Using Enums improves both validation and autocomplete in documentation/UI.

mehalyna, line 1: You might want to support date ranges or number ranges in the future. It could be good to define:
```
date_to: Optional[date] = Field(None, description="Filter until this date")
```
Similarly, for sorting/filters like `views_count` or `likes_count`, you could consider `min_likes`, `max_views`, etc.

mehalyna, line 1: You could add a validator to ensure `sort_by` contains only allowed fields (to avoid user tampering):
```
from pydantic import root_validator

class BlogPostSearchParams(PaginationAndSortingMixin):
    ...
    sort_by: Optional[str] = Field("likes_count")

    @root_validator
    def validate_sort_by(cls, values):
        allowed_fields = {"likes_count", "created_at", "title"}
        sort_field = values.get("sort_by")
        if sort_field not in allowed_fields:
            raise ValueError(f"Invalid sort_by field: {sort_field}")
        return values
```

BackEnd/search-service/app/search_processing/search_processor.py
```diff
@@ -0,0 +1,178 @@
...
+            index_name=self.index_name,
+            query=params.query,
+            filters=filters,
+            sort_by=params.sort_by,
+            search_fields=self.search_fields,
+            sort_order=params.sort_order,
+            page=params.page,
+            page_size=params.page_size,
+            include_suggestions=params.include_suggestions,
+            suggest_filed=self.suggest_filed
```
mehalyna: Fix typo  → `suggest_field`

BackEnd/search-service/app/search_processing/search_query.py
```diff
@@ -0,0 +1,82 @@
...
+        query: Optional[str],
+        filters: List[Q],
+        sort_by: Optional[str],
+        search_fields: List[str],
+        sort_order: str,
+        page: int,
+        page_size: int,
+        include_suggestions: bool,
+        suggest_filed: str
+) -> (int, List[Dict[str, Any]]):
```
mehalyna: Function signature says it returns a tuple of two values. 
But it actually returns three:
```
return total_records, records, response
```

mehalyna (changes requested):

BackEnd/search-service/app/routes/search.py
NikitaPash: Thanks for the remark. Right now I don't see a need to implement this. Search is working with mock data and I can't be sure what data will be searched by users in the future

BackEnd/search-service/app/search_processing/local_search_params.py
NikitaPash: Added date ranges for events, but not sure about likes and views

mehalyna (approved):

hrybvalerii (approved):
Looks good to me

AntonKovalenk0 (approved):
All Great!

PR #193, opened 2025-03-30
[News] Task#170 add filtering

Maksym04u:
- Implemented keyword-based filtering to filter out irrelevant news.
- Wrote README file


boghtml (approved):

BackEnd/news-service/app/filtering.py
```diff
@@ -0,0 +1,37 @@
...
+    matched_categories = set()
+
+    for category, pattern in compiled_patterns.items():
+        if pattern.search(article_text) or pattern.search(article_title):
+            matched_categories.add(category)  
+
+        if len(matched_categories) >= 3:
+            return True
+
+    return False
```
mehalyna: Make your function more reusable with a configurable threshold:
```
def is_business_news(article_text: str, article_title: str, min_categories: int = 3) -> bool:
    ...
    if len(matched_categories) >= min_categories:
        return True
```

BackEnd/news-service/app/filtering.py
mehalyna, line 1: If you want to classify or label the article later, you can return the set too:
```
def get_business_news_categories(article_text: str, article_title: str) -> set:
    return {
        category for category, pattern in compiled_patterns.items()
        if pattern.search(article_text) or pattern.search(article_title)
    }
```

mehalyna (changes requested):

hrybvalerii (approved):
Good job!

mehalyna (approved):

PR #203, opened 2025-04-05
[Authentication] Task#202 Kafka for Authentication and Profiles

sofiahnatiuk:
1. Updated registration flow, allowing users to register 2 companies (both investor and startup) or just one company
2. Updates registration tests
3. Added a Kafka producer to notify Profiles about a recently registered user, who needs a profile
4. Added a Kafka consumer to receive information about user role updates from Profiles (Validated -> NotValidated or NotValidated -> Validated)

### DIFF
--- a/BackEnd/authentication_service/custom_auth/consumers.py
+++ b/BackEnd/authentication_service/custom_auth/consumers.py
@@ -1,0 +1,62 @@
+import json
+import logging
+from kafka import KafkaConsumer
+from django.conf import settings
+from .models import CustomUser, Role, UserRole
+
+logger = logging.getLogger(__name__)
+
+KAFKA_BROKER = getattr(settings, "KAFKA_BROKER", "kafka:9092")
+USER_ROLE_UPDATE_TOPIC = "user_role_update"
+
+consumer = KafkaConsumer(
+    USER_ROLE_UPDATE_TOPIC,
+    bootstrap_servers=KAFKA_BROKER,
+    value_deserializer=lambda m: json.loads(m.decode("utf-8")),
+    group_id="user-role-update-group",
+)
+
+def process_role_update_message(message):
+    """
+    Process incoming role update message and update user's role status in the database.
+
+    Args:
+        message (dict): The message containing user_id, profile_type, and status.
+    """
+    try:
+        user_id = message.get("user_id")
+        profile_type = message.get("profile_type")
+        new_status = message.get("status")
+
+        if not user_id or not profile_type or new_status not in ["validated", "not_validated"]:
+            logger.error(f"Invalid message: {message}")
+            return
+
+        user = CustomUser.objects.get(id=user_id)
+
+        role = Role.objects.get(name=profile_type.capitalize())  # Capitalize to match 'Startup' or 'Investor'
+
+        user_role = UserRole.objects.get(user=user, role=role)
+
+        user_role.status = new_status
+        user_role.save()
+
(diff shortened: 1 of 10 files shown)


AntonKovalenk0 (approved):

isac-medison (approved):

hrybvalerii (approved):

BackEnd/authentication_service/custom_auth/consumers.py
VovaD199, line 1: Not integrated into Django lifecycle

Currently this code simply calls process_role_update_message() when importing a module - this is not the Django way.

Convert the logic to a Django management command.

This will allow you to run kafka_consumer via python manage.py, easily log, scale, stop, manage via supervisor/systemd, etc.

VovaD199, line 1: No retry/backoff on KafkaException
When a Kafka error occurs (e.g. broker unavailability), the script immediately throws an exception and stops.

Use exponential backoff + logging

VovaD199, line 1: while True without controlled stop
The service can only be stopped manually (Ctrl+C). In production environments it is better to have control (via a flag or signal).

Use the self.running variable or threading.Event().
Or listen to system signals (signal.signal(signal.SIGINT, ...)).

VovaD199, line 1: Insufficient JSON message validation
JSON may not contain the user_id, profile_type, or status keys, and then processing will break or be incorrect (e.g., None in get()).

BackEnd/authentication_service/custom_auth/producers.py
```diff
@@ -33,17 +36,48 @@ def send_message(message_type: str, email: str, link: str, name: str = None):
...
+        # Produce message to Kafka
+        auth_producer.produce(AUTH_TOPIC, value=json.dumps(message))
+        auth_producer.flush()
+        logger.info(f"Message sent to {AUTH_TOPIC}")
     except Exception as e:
         logger.error(f"Failed to send message: {e}")
 
-def close_producer():
-    """Closes the Kafka producer gracefully."""
+
```
VovaD199, line 44: Using flush() after each message
flush() blocks the thread while waiting for delivery confirmation. This slows down performance, especially with a large number of messages.

It is better to use a delivery callback or flush() only at the end of the session:

def delivery_callback(err, msg):
    if err:
        logger.error(f"Delivery failed: {err}")
    else:
        logger.info(f"Message delivered to {msg.topic()} [{msg.partition()}]")

auth_producer.produce(AUTH_TOPIC, value=json.dumps(message), callback=delivery_callback)

VovaD199 (changes requested):

BackEnd/authentication_service/custom_auth/consumers.py
sofiahnatiuk: I already have a Django management command in management/commands/consume_roles.py

sofiahnatiuk: Added exponential backoff and more logging

sofiahnatiuk: Added checks and logs for data missing from JSON

BackEnd/authentication_service/custom_auth/producers.py
sofiahnatiuk: Added delivery callback

VovaD199 (approved):

PR #204, opened 2025-04-06
Notification service

isac-medison:
[add] Paralel execution of two topics (profile_image and auth)


hrybvalerii (approved):

IraMladanovych02 (approved):

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
```diff
@@ -30,8 +31,10 @@ object EmailConstants {
 
   object EmailTemplates {
     final lazy val TEMPLATES_PATH: String = sys.env.getOrElse("EMAIL_TEMPLATES_PATH", "src/main/scala/com/softserve/email_templates") 
-    final lazy val CONFIRM_NAME: String = "ConfirmEmailTemplate.mustache"
+    final lazy val AUTH_CONFIRM: String = "ConfirmEmailTemplate.mustache"
     final lazy val EVENT_NAME: String = "EventActionTemplate.mustache"
```
VovaD199: The name of the constant EVENT_NAME does not match its value.

EVENT_NAME looks like an event name, but contains a letter template. It should be renamed:
final lazy val EVENT_ACTION: String = "EventActionTemplate.mustache"

BackEnd/notifications-service/src/main/scala/com/softserve/consumers/MessageConsumer.scala
```diff
@@ -0,0 +1,131 @@
...
+    templateEngine: TemplateEngine
+  ): ZIO[Any, Throwable, Unit] = {
+    json.fromJson[T] match {
+      case Left(error) =>
+        printLine(s"[ERROR] Failed to parse JSON: $error")
+
+      case Right(payload) =>
+        val errors = validate(payload)
+        if (errors.nonEmpty)
+          ZIO.debug(s"Validation failed: ${errors.mkString(", ")}")
```
VovaD199: Now, if validation fails:
ZIO.debug(...) // but doesn't stop the thread explicitly.
You can do this:
ZIO.logWarning(s"Validation failed: ${errors.mkString(", ")}") *> ZIO.unit

BackEnd/notifications-service/src/main/scala/com/softserve/consumers/MessageConsumer.scala
```diff
@@ -0,0 +1,131 @@
...
+          SmtpEmailSender.sendEmail(
+            emailExtractor(payload),
+            subject,
+            emailBody
+          ).provideLayer(EmailConstants.emailConfigLayer).unit
+        }
+    }
+  }
+
+  private def getAuthTemplateName(auth: AuthMessage): String =
```
VovaD199: getAuthTemplateName has a "magic string" → "unknown".
This is dangerous - it is better to return Option[String] and handle None, or call ZIO.fail:
case other => throw new IllegalArgumentException(s"Unknown auth message_type: $other")

BackEnd/notifications-service/src/main/scala/com/softserve/consumers/MessageConsumer.scala
VovaD199, line 1: Topic processing is strictly defined
If there are more topics, you will have to change match-case. You can replace it with Map:
val topicHandlers: Map[String, (CommittableRecord[String, String]) => ZIO[Any, Throwable, Unit]] = Map(
  TopicConstants.AUTHENTICATION -> (record => ...),
  TopicConstants.PROFILE_IMAGE -> (record => ...)
)

BackEnd/notifications-service/src/main/scala/com/softserve/models/ProfielMessage.scala
```diff
@@ -0,0 +1,44 @@
...
+      ValidationUtils.validateString(updated_at, "Updated at") ++
+      ValidationUtils.validateString(moderation_time, "Moderation time") ++
+      ValidationUtils.validateString(image_path, "Image path") ++
+      ValidationUtils.validateString(profile_view_url, "Profile view url")
+
+    if (errors.nonEmpty) Unsafe.unsafe { implicit u =>
+      Runtime.default.unsafe.run(ZIO.logError(s"Validation failed: ${errors.mkString(", ")}"))
+    } 
+    errors
+  }
```
VovaD199, line 23: Avoid using Unsafe.unsafe in validate.
Why this is bad:
Calling unsafe.run blocks execution.
This goes against the reactive nature of ZIO.
Unitary methods like validate are better made pure functions.
Option 1 (better): move logging outside:
def validate: List[ValidationUtils.ValidationError] = {
  ValidationUtils.validateEmail(email) ++
    ValidationUtils.validateString(profile_name, "Profile name") ++
    ValidationUtils.validateString(updated_at, "Updated at") ++
    ValidationUtils.validateString(moderation_time, "Moderation time") ++
    ValidationUtils.validateString(image_path, "Image path") ++
    ValidationUtils.validateString(profile_view_url, "Profile view url")
}

And log in processing:
val errors = payload.validate
if (errors.nonEmpty)
  ZIO.logError(s"Validation failed: ${errors.mkString(", ")}")

BackEnd/notifications-service/src/main/scala/com/softserve/producers/AuthMessageProducer.scala
```diff
@@ -7,27 +7,42 @@ import com.softserve.serde._
...
-          AuthMessageSerde.serde
-          ).retry(retryStrategy)
+          key,
+          message,
+          keySerde,
+          valueSerde
+        ).retry(retryStrategy)
       } yield ()
     }.mapError { e =>
       Console.printLine(s"Error producing message to Kafka: ${e.getMessage}")
```
VovaD199: mapError at the end returns nothing
This looks like an error — Console.printLine(...) returns ZIO, and mapError expects a function that returns Throwable.

That is, this expression is not typed correctly, and it:

either does not compile,

or does not do what is expected.

Option 1: correct error logging:
.tapError(e => Console.printLine(s"Error producing message to Kafka: ${e.getMessage}"))

BackEnd/notifications-service/src/main/scala/com/softserve/producers/AuthMessageProducer.scala
VovaD199, line 1: Currently, only produce(...) is retried. If the Producer.make(...) creation fails (for example, Kafka is temporarily unavailable), retry will not work.
You can retry the entire operation within ZIO.scoped { ... }:
val scopedProduce = ZIO.scoped {
  for {
    producer <- Producer.make(producerSettings)
    _ <- producer.produce(...)
  } yield ()
}

scopedProduce.retry(retryStrategy)

BackEnd/notifications-service/src/main/scala/com/softserve/serde/ProfielMessageSerde.scala
```diff
@@ -0,0 +1,16 @@
...
+import zio._
+import zio.json._
+import zio.kafka.serde.Serde
+import com.softserve.models._
+
+object ProfileMessageSerde {
+  val serde: Serde[Any, ProfileMessage] =
+    Serde.string.inmap(
+      _.fromJson[ProfileMessage].getOrElse {
+        println("Invalid Profile JSON, using fallback.")
```
VovaD199, line 12: println(...) is an antipattern in ZIO
Using println in serialization is an antipattern in ZIO projects. It:

Violates the principles of pure FP;

Logs poorly (especially with large volumes);

Cannot be configured/managed (unlike ZIO logging).

Solution: use ZIO logging (ZIO.logWarning), but since inmap is a synchronous operation, you will either have to switch to inmapZIO, or do the logging at a higher level (for example, in processMessage).

BackEnd/notifications-service/src/main/scala/com/softserve/serde/ProfielMessageSerde.scala
```diff
@@ -0,0 +1,16 @@
...
+        ProfileMessage.empty
```
VovaD199, line 13: fallback on .empty can be dangerous
If the JSON is invalid, we create a valid but empty object.
Then it will most likely fail validation.
But the process(...) logic in MessageConsumer.scala already checks .validate, and if there are errors, they are logged, but the process continues.
This may be ok if:
It is intended that way (e.g., to avoid crashing the stream),
But then the logging should be explicitly visible, not just println.

BackEnd/notifications-service/src/main/scala/com/softserve/serde/ProfielMessageSerde.scala
VovaD199: Alternative: use Serde.string.inmapZIO with a valid log:
val serde: Serde[Any, ProfileMessage] = Serde.string.inmapZIO { str =>
  str.fromJson[ProfileMessage] match {
    case Right(p) => ZIO.succeed(p)
    case Left(err) =>
      ZIO.logWarning(s"Invalid Profile JSON: $err").as(ProfileMessage.empty)
  }
}(p => ZIO.succeed(p.toJson))

BackEnd/notifications-service/src/main/scala/com/softserve/tests/test-data/MockData.scala
```diff
@@ -2,10 +2,24 @@ package com.softserve.testData
 import com.softserve.models._
 
 object MockData {
-  final lazy val auth_activation_message = AuthMessage(
+  final lazy val auth_activation = AuthMessage(
```
VovaD199: It's better to stick to the general Scala style, for examlpe:
final lazy val authActivation
final lazy val authPasswordChange
final lazy val profileImage

BackEnd/notifications-service/src/main/scala/com/softserve/tests/test-data/MockData.scala
```diff
@@ -2,10 +2,24 @@ package com.softserve.testData
...
+      email = "email.example@gmail.com",
+      message_type = "password-reset",
+      name = "User#1",
+      link = "http://example.com"
+  )
+  final lazy val profile_image = ProfileMessage(
+      email = "email.example@gmail.com",
+      profile_name = "Mock Name",
+      updated_at = "11.10.2003",
+      moderation_time = "11.10.2003",
```
VovaD199: It is better to make the data in ISO format.
If the rest of the system expects a different format, that's fine. But for most cases, ISO is better.
for example:
    updated_at = "2003-10-11",
    moderation_time = "2003-10-11",

VovaD199 (changes requested):

BackEnd/notifications-service/src/main/scala/com/softserve/constants/EmailConstants.scala
isac-medison: This constant will not be used, so I am removing it and its template

BackEnd/notifications-service/src/main/scala/com/softserve/models/ProfielMessage.scala
isac-medison: Thanks for finding that, I did not see it, but for some reason I am validating it two times (in validator and when I call that validator)

BackEnd/notifications-service/src/main/scala/com/softserve/producers/AuthMessageProducer.scala
isac-medison: Solid advice, I decided to use     .catchAll { err =>
      Console.printLine(s"Error producing message to Kafka: ${err.getMessage}") *> ZIO.unit
    }
   in order to catch all errors and continue execution

VovaD199 (approved):

